PR #16, opened 2017-03-23
PING/PONG

### LINKED ISSUE #12: support PING-PONG

newellista:
This will reply to server `PING` messages.

This is the first iteration.  There is another PR with a different solution.

PROS:

  * Simpler parser
  * Message isn't "misrepresented" from the parser.  It returns just the `PING`, not manufacturing fake `sid` or `body`

CONS:
  * Handler is a little funky.  Since the message is a different format, it requires a bit more work in the handler to deal with the differences.

Resolves #12

/cc @newellista @tallguy-hackett @jjcarstens

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -40,7 +40,7 @@ defmodule Gnat do
   defp process_message({:msg, topic, sid, body}, state) do
     send state.receivers[sid], {:msg, topic, body}
   end
-  defp process_message("PING", state) do
+  defp process_message(:ping, state) do
     :gen_tcp.send(state.tcp, "PONG\r\n")
   end
   def handle_info({:tcp, tcp, data}, %{tcp: tcp, parser: parser}=state) do
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -37,11 +37,16 @@ defmodule Gnat do
     end
   end
 
+  defp process_message({:msg, topic, sid, body}, state) do
+    send state.receivers[sid], {:msg, topic, body}
+  end
+  defp process_message("PING", state) do
+    :gen_tcp.send(state.tcp, "PONG\r\n")
+  end
   def handle_info({:tcp, tcp, data}, %{tcp: tcp, parser: parser}=state) do
+    Logger.debug "#{__MODULE__} received #{inspect data}"
     {new_parser, messages} = Parser.parse(parser, data)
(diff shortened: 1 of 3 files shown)


lib/gnat.ex
```diff
@@ -37,11 +37,16 @@ defmodule Gnat do
     end
   end
 
+  defp process_message({:msg, topic, sid, body}, state) do
```
mmmries: I definitely like this function with pattern matching more than a `case` statement used in your other PR

mmmries:
Overall I prefer this PR to #17

Getting back different shapes of data for `PING` vs a `MSG` makes more sense to me since they are different entities as far as the nats protocol goes.

mmmries:
Once we get started on partial message parsing we should also handle the case where someone publishing a message like:

```
MSG 1 4\r\n
PING\r\n
```

In that case the `PING` is the body of a message not a protocol command

newellista:
Interesting.  I was leaning more toward the solution in #17, but I'm ok with this one.

I think the case where `PING` is part of the body, can be handled by a guard clause ensuring that the partial message is empty.  I can't imagine a scenario where the partial message contains the end of a message, followed immediately by a `PING\r\n` command, so I don't think we'll have to deal with that case.

I'll go ahead and close #17.

mmmries:
If you prefer #17 maybe we should get a 3rd opinion?

quixoten:
Maybe it's too early to make this change, or perhaps it doesn't fit with elixir, but could we register callbacks in the parser when it's first created? 

```
Parser.on_msg(ref_to_process_msg_def)
parser.on_ping(ref_to_process_ping_def)
```

lib/gnat.ex
```diff
@@ -37,11 +37,16 @@ defmodule Gnat do
...
+    send state.receivers[sid], {:msg, topic, body}
+  end
+  defp process_message("PING", state) do
```
quixoten: Why `"PING"` and not `:ping` or even `{:ping}`. Am I worried to much about symmetry?

mmmries:
I don't love the idea of callbacks. I think the strong tendency in functional languages is to return data unless you need to capture closures or do something else with a function.

:+1: for `:ping` as a the best way to represent it, that is the easiest thing to pattern match on

mmmries:
:shipit: this is a good first cut

mmmries:
:shipit: :shipit: :shipit: :shipit: :shipit:

PR #17, opened 2017-03-23
PING/PONG -- Take #2

### LINKED ISSUE #12: support PING-PONG

newellista:
This is the second implementation handling PING/PONG.

It feels a little cleaner, and the code is a bit simpler.

EXCEPT:

The parser generates an "artificial" message containing empty `sid` and `body` fields.  Probably not a big deal, but looks a little strange.

Anybody have any feedback on a possible **third** solution that is more idiomatic?

Resolves #12 

/cc @newellista @tallguy-hackett @jjcarstens

### DIFF
--- a/lib/gnat/parser.ex
+++ b/lib/gnat/parser.ex
@@ -12,6 +12,7 @@ defmodule Gnat.Parser do
   end
 
   def parse(parser, "", parsed), do: {parser, Enum.reverse(parsed)}
+  def parse(parser, "PING\r\n", _), do: {parser, [{:msg, "PING", "", ""}]}
   def parse(parser, bytes, parsed) do
     {index, 2} = :binary.match(bytes, "\r\n")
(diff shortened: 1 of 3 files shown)


newellista:
Closing in favor of solution in #16.

lib/gnat/parser.ex
```diff
@@ -12,6 +12,7 @@ defmodule Gnat.Parser do
   end
 
   def parse(parser, "", parsed), do: {parser, Enum.reverse(parsed)}
+  def parse(parser, "PING\r\n", _), do: {parser, [{:msg, "PING", "", ""}]}
```
quixoten: Overall I found this version more readable, but this is also the first elixir project I've read; however, I wonder if exposing PING as a topic is hiding a better solution. In the protocol, PING and MSG represent the same idea, but right here we're forcing PING to be represent a topic/subject, which is not the same idea that PING represents in the protocol.

PR #19, opened 2017-03-24
Unsub with max msgs

mmmries:
Getting a little further with #8 

This PR adds support for doing an `UNSUB` with a maximum number of messages. I decided to make the maximum messages be part of an optional 3rd parameter which comes in as a keyword list. This seems to match a common pattern for APIs in the elixir community. Right now this is the only optional parameter to an unsub command in the NATS protocol, but I imagine that could change in the future so I figured modeling it as a keyword list kept things flexible for the future.

I also didn't want to add multiple `handle_call` clauses or private functions to the `gnat.ex` so I decided to extract the formatting of on-the-wire commands into a separate module.

### DIFF
--- a/lib/gnat/command.ex
+++ b/lib/gnat/command.ex
@@ -1,0 +1,7 @@
+defmodule Gnat.Command do
+  @newline "\r\n"
+  @unsub "UNSUB"
+
+  def build(:unsub, sid, []), do: [@unsub, " #{sid}", @newline]
+  def build(:unsub, sid, [max_messages: max]), do: [@unsub, " #{sid}", " #{max}", @newline]
+end
--- a/test/command_test.exs
+++ b/test/command_test.exs
@@ -1,0 +1,14 @@
+defmodule Gnat.CommandTest do
+  use ExUnit.Case, async: true
+  alias Gnat.Command
+
+  test "formatting a simple unsub message" do
+    command = Command.build(:unsub, 12, []) |> IO.iodata_to_binary
+    assert command == "UNSUB 12\r\n"
+  end
+
+  test "formatting an unsub message with max messages" do
+    command = Command.build(:unsub, 12, [max_messages: 3]) |> IO.iodata_to_binary
+    assert command == "UNSUB 12 3\r\n"
+  end
+end
--- a/test/gnat_test.exs
+++ b/test/gnat_test.exs
@@ -43,4 +43,20 @@ defmodule GnatTest do
       after 200 -> :ok
     end
   end
+
+  test "unsubscribing from a topic after a maximum number of messages" do
+    topic = "testunsub_maxmsg"
+    {:ok, pid} = Gnat.start_link()
+    {:ok, sub_ref} = Gnat.sub(pid, self(), topic)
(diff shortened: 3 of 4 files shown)


test/command_test.exs
```diff
@@ -0,0 +1,14 @@
...
+  use ExUnit.Case, async: true
+  alias Gnat.Command
+
+  test "formatting a simple unsub message" do
+    command = Command.build(:unsub, 12, []) |> IO.iodata_to_binary
+    assert command == "UNSUB 12\r\n"
+  end
+
+  test "formatting an unsub message with max messages" do
+    command = Command.build(:unsub, 12, [max_messages: 3]) |> IO.iodata_to_binary
```
quixoten: Your use if iodata_to_binary suggests that "iodata" and arrays (or lists in elixir parlance!) are the same thing. Is that true?

lib/gnat/command.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Gnat.Command do
+  @newline "\r\n"
+  @unsub "UNSUB"
+
+  def build(:unsub, sid, []), do: [@unsub, " #{sid}", @newline]
+  def build(:unsub, sid, [max_messages: max]), do: [@unsub, " #{sid}", " #{max}", @newline]
```
quixoten: The matching in method signatures is pretty awesome! I like having the commands in a common place.

test/gnat_test.exs
```diff
@@ -43,4 +43,20 @@ defmodule GnatTest do
...
+
+  test "unsubscribing from a topic after a maximum number of messages" do
+    topic = "testunsub_maxmsg"
+    {:ok, pid} = Gnat.start_link()
+    {:ok, sub_ref} = Gnat.sub(pid, self(), topic)
+    :ok = Gnat.unsub(pid, sub_ref, max_messages: 2)
+    :ok = Gnat.pub(pid, topic, "msg1")
+    :ok = Gnat.pub(pid, topic, "msg2")
+    :ok = Gnat.pub(pid, topic, "msg3")
+    assert_receive {:msg, ^topic, "msg1"}, 500
```
quixoten: How does this assert_receive magic work?

test/command_test.exs
quixoten: That ! was meant to be a ?

mmmries: In erlang there is a generally accepted type called `iodata` (see [types and specs](http://erlang.org/doc/reference_manual/typespec.html)) which most network calls or other IO related functions accept. It can be a nested list of lists that contain binaries or character lists.

For simplicity in my test I am just converting back to binary so I can do a simple equality match, but we would actually pass this to the network as a list of small strings. It avoids allocating additional strings by the `UNSUB` and `\r\n` parts of the string will always be there so we only dynamically allocate the parts of the string that need to be dynamic.

test/gnat_test.exs
mmmries: [It's build into exunit](https://hexdocs.pm/ex_unit/ExUnit.Assertions.html#assert_receive/3), part of the standard library. It basically asserts that within 500 milliseconds this process will receive a message that matches the pattern

Issue #10, opened 2017-03-23
Add a RPC style helper that does a SUB -> UNSUB -> PUB and then waits for the reply


mmmries:

mmmries:
Addressed in #22

PR #22, opened 2017-03-24
Pub with reply

mmmries:
Support for passing a `reply_to` and receiving messages with a `reply_to` value. This is a pre-requisite for #10

In order to support this I needed to change the format of how we deliver messages to subscribed processes. Instead of passing another element in a tuple (I always forget which order the items are in) I changed it to passing a map so you can pattern match in a more readable way.

This is a YOLO breaking change, but we haven't even published to hex.pm so :man_shrugging: :woman_shrugging: 

/cc @newellista @quixoten @film42 @tallguy-hackett


newellista:
:shipit:

PR #23, opened 2017-03-24
Request reply

mmmries:
This introduces a helper function to make doing [request and reply](http://nats.io/documentation/concepts/nats-req-rep/) patterns easier and more efficient (ie by sending the sub, unsub and pub messages together in the same TCP packet).

I also updated the benchmark script and I currently get these numbers on my machine:

```
$ mix run bench.exs
Erlang/OTP 19 [erts-8.2] [source] [64-bit] [smp:8:8] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
Elixir 1.4.1
Benchmark suite executing with the following configuration:
warmup: 2.0s
time: 10.0s
parallel: 1
inputs: none specified
Estimated total run time: 48.0s

Benchmarking parse-128...
Benchmarking pub - 128...
Benchmarking req-reply-16...
Benchmarking sub-unsub-pub-16...

Name                       ips        average  deviation         median
pub - 128              89.12 K       11.22 μs   ±117.84%       11.00 μs
parse-128              89.06 K       11.23 μs   ±173.47%       10.00 μs
sub-unsub-pub-16        8.55 K      116.96 μs    ±28.81%      111.00 μs
req-reply-16            5.83 K      171.55 μs    ±18.23%      165.00 μs
```

Doing 5,830 request-reply cycles in serial per second is 👌  in my book. We'll see how close we can stay to those numbers as we add more complete parsing and behavior.

/cc @film42 @newellista @jjcarstens @tallguy-hackett

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -24,6 +24,13 @@ defmodule Gnat do
 
   def pub(pid, topic, message, opts \\ []), do: GenServer.call(pid, {:pub, topic, message, opts})
 
+  def request(pid, topic, body, opts \\ []) do
+    recipient = Keyword.get(opts, :recipient, self())
+    max_messages = Keyword.get(opts, :max_messages, 1)
+    inbox = "INBOX-#{:crypto.strong_rand_bytes(12) |> Base.encode64}"
+    GenServer.call(pid, {:request, %{recipient: recipient, max_messages: max_messages, inbox: inbox, body: body, topic: topic}})
+  end
+
   @doc """
   Unsubscribe from a topic
 
@@ -87,6 +94,15 @@ defmodule Gnat do
     command = Command.build(:pub, topic, message, opts)
     :ok = :gen_tcp.send(state.tcp, command)
     {:reply, :ok, state}
+  end
+  def handle_call({:request, request}, _from, %{next_sid: sid}=state) do
+    sub = ["SUB", " #{request.inbox} #{sid}", "\r\n"]
+    unsub = Command.build(:unsub, sid, [max_messages: request.max_messages])
+    pub = Command.build(:pub, request.topic, request.body, reply_to: request.inbox)
+    receivers = Map.put(state.receivers, sid, request.recipient)
+    next_sid = sid + 1
+    :ok = :gen_tcp.send(state.tcp, [sub, unsub, pub])
+    {:reply, {:ok, request.inbox}, %{state | receivers: receivers, next_sid: next_sid}}
   end
   def handle_call({:unsub, sid, opts}, _from, state) do
     command = Command.build(:unsub, sid, opts)
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -25,20 +25,33 @@ defmodule Gnat do
   def pub(pid, topic, message, opts \\ []), do: GenServer.call(pid, {:pub, topic, message, opts})
 
   @doc """
-  Send a request listen for response(s)
+  Send a request and listen for a response synchronously
 
   Following the nats [request-response pattern](http://nats.io/documentation/concepts/nats-req-rep/) this
   function generates a one-time topic to receive replies and then sends a message to the provided topic.
 
   Supported options:
-    * max_messages: how many messages to expect before cleaning up the inbox subscription (default `1`)
-    * recipient: which pid should receive the responses (default `self()`)
+    * receive_timeout: an integer number of milliseconds to wait for a response. Defaults to 60_000
+
+  ```
+  {:ok, gnat} = Gnat.start_link()
+  case Gnat.request("i_can_haz_cheezburger", "plZZZZ?!?!?") do
+    {:ok, %{body: delicious_cheezburger}} -> :yum
+    {:error, :timeout} -> :sad_cat
+  end
+  ```
   """
   def request(pid, topic, body, opts \\ []) do
-    recipient = Keyword.get(opts, :recipient, self())
-    max_messages = Keyword.get(opts, :max_messages, 1)
+    receive_timeout = Keyword.get(opts, :receive_timeout, 60_000)
     inbox = "INBOX-#{:crypto.strong_rand_bytes(12) |> Base.encode64}"
-    GenServer.call(pid, {:request, %{recipient: recipient, max_messages: max_messages, inbox: inbox, body: body, topic: topic}})
+    {:ok, subscription} = GenServer.call(pid, {:request, %{recipient: self(), inbox: inbox, body: body, topic: topic}})
+    receive do
+      {:msg, %{topic: ^inbox}=msg} -> {:ok, msg}
+      other -> IO.inspect(other)
+      after receive_timeout ->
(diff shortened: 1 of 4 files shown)


mmmries:
Derp, just remembered that I need to document the new `request` function. I'll push up another commit, but interested in feedback in the meantime.

film42:
Looks good! Do you want to add support for a synchronous req/ reply?

mmmries:
@film42 since our plan for MX is to have a reply that gets back 2 messages (ack + response with different timeouts) it seemed like a synchronous req-reply wouldn't be that useful. But maybe it is useful in a broader sense?  We could add the `receive do` into the `request` function to make it synchronous

mmmries:
```elixir
defmodule DoubleReplyClient do
  def request(gnat, topic, body) do
    {:ok, inbox} = Gnat.request(gnat, topic, body, max_messages: 2)
    receive do
      {:msg, %{topic: ^inbox, body: <<1>>}} ->
        receive do
          {:msg, %{topic: ^inbox, body: response}} ->
            {:ok, response}
          after 30_000 -> {:error, :response_timeout}
        end
      after 500 -> {:error, :ack_timeout}
    end
  end
end
```

Here is a double-response synchronous request function. Did a quick benchmark and got 4k serial requests per second.

mmmries:
@film42 do you have a plan for dealing with the case where you timeout waiting for the ack and the subscription doesn't clean up after itself? Are you going to call an additional `unsub` in those cases to avoid leaving a bunch of inboxes hanging around?

film42:
I'm cleaning those up here: https://github.com/abrandoned/protobuf-nats/blob/d93d62df7761a22b446426565b2ac8109b0e7433/lib/protobuf/nats/client.rb#L83-L85

However I wouldn't consider the double reply to be something this lib should handle. If you expect to receive more than one reply, then you should be using `subscribe` over `request`, because `request` isn't actually part of the protocol, it's a convention built with the protocol to get a response from some other subscriber.

In the go client, they always clean up the inbox subscription create by `request`: https://github.com/nats-io/go-nats/blob/399fe00b07f048bdacbdbfdfeda39712642e6a11/nats.go#L1909-L1928 To do this they `subscribe`, tell the broker to unsub after 1 message, and explicitly unsub after the a response message arrives or timeout occurs. They're keeping track of the subscriptions though, so no message is sent if the subscription is already removed.

In their pure ruby lib they allow for async requests, where the max response can be more than one message, but this is basically passing your options to `subscribe` and then a call to `publish` for you. This is ok, but they're not attempting to clean anything up for you. If I was maintaining the lib, I would scrap the async request behavior.

There's an interesting discussion here about adding async req/rep to cnats, where a major concern is having to rely on the subscription getting a message to clean up mem. Their solution is to not make this a concern of req/rep but to allow a timer to be added to a subscription so the lib can clean it up for you. Interesting trade-off here.

IMO making `request` a "blocking" library feature that cleans up the subscription in all cases and only allows one message is the way to go. Since this is what the go-nats client is doing, I would consider it "expected behavior." If you need something beyond that, like we do in the MX pb nats ruby lib, then it's easy to write/ test your own protocol.

mmmries:
@film42 I think you're right that it probably makes sense to implement `request` such that it blocks waiting for a single reply to follow the pattern in the go client.

I was thinking that the main reason I wanted to support a `request` call is that it can send the `SUB`, `UNSUB` and `PUB` messages all in a single TCP packet (as long as the request body is under ~1k).

Since performance is one of my goals, maybe I should make two functions. `Gnat.request` will synchronously wait for a single reply with a configurable timeout. `Gnat.async_request` will support a configurable number of replies and won't be responsible for cleaning up the subscription. It would mostly be a convenience + performance function at that point.

film42:
@mmmries 👍 That sounds good. It might be helpful to expose that behavior, though. Some way to tell the lib that you want to fit N messages in as few TCP packets as possible. The idea of a message group. But I think the `async_request` is good enough for the optimization and behavior that we want.

mmmries:
I decided to validate my idea that doing the `sub` + `unsub` + `pub` all in one tcp action actually makes a difference and found that the difference is actually very small:

The `sub-unsub-pub-wait-16` is doing the same thing as the `req-reply` except that it is sending the commands to the genserver individually rather than sending a single message. T
```
Name                            ips        average  deviation         median
req-reply-16                 5.86 K      170.62 μs    ±13.31%      165.00 μs
sub-unsub-pub-wait-16        5.20 K      192.20 μs    ±19.71%      188.00 μs
```

I rant the benchmark several times and `req-reply` is always slightly faster, but never by more than 500 ips (less than 10% different).

So I'll go ahead and just implement the `request` as a synchronous method that cleans up its own subscription for conveninence, and leave the `async_request` out for now since it doesn't really improve performance.

film42:
Oh interesting. With such a high std deviation, I think that's a good call. :+1:

mmmries:
Updated to make `request` synchronous, cleanup its own subscription in case of timeout and have an optional `receive_timeout` (default 60sec). Here are the latest benchmark results:

```
Elixir 1.4.1
Benchmark suite executing with the following configuration:
warmup: 2.0s
time: 10.0s
parallel: 1
inputs: none specified
Estimated total run time: 48.0s

Benchmarking parse-128...
Benchmarking pub - 128...
Benchmarking req-reply-4...
Benchmarking sub-unsub-pub-16...

Name                       ips        average  deviation         median
parse-128              91.33 K       10.95 μs   ±183.99%       10.00 μs
pub - 128              80.54 K       12.42 μs   ±137.45%       10.00 μs
sub-unsub-pub-16        7.70 K      129.79 μs    ±41.43%      111.00 μs
req-reply-4             5.21 K      191.81 μs    ±31.28%      167.00 μs
```

lib/gnat.ex
```diff
@@ -25,20 +25,33 @@ defmodule Gnat do
...
   def request(pid, topic, body, opts \\ []) do
-    recipient = Keyword.get(opts, :recipient, self())
-    max_messages = Keyword.get(opts, :max_messages, 1)
+    receive_timeout = Keyword.get(opts, :receive_timeout, 60_000)
     inbox = "INBOX-#{:crypto.strong_rand_bytes(12) |> Base.encode64}"
-    GenServer.call(pid, {:request, %{recipient: recipient, max_messages: max_messages, inbox: inbox, body: body, topic: topic}})
+    {:ok, subscription} = GenServer.call(pid, {:request, %{recipient: self(), inbox: inbox, body: body, topic: topic}})
+    receive do
+      {:msg, %{topic: ^inbox}=msg} -> {:ok, msg}
+      other -> IO.inspect(other)
```
film42: Do you ever expect this clause to be hit? If so, do we need to unsub here as well?

mmmries: Derp, I put this in when debugging. I don't expect it to be hit, but if anyone ever sent us a message while we were waiting for a response it would trigger this clause and could mess up some other piece of code that was expecting a message. I removed this in the latest commit

Issue #18, opened 2017-03-24
Handle protocol operation names in a case insensitive manner


tallguy-hackett:
According to the [NATS protocol documentation](http://nats.io/documentation/internals/nats-protocol/), protocol operation names are case insensitive

> NATS protocol operation names are case insensitive, thus `SUB foo 1\r\n` and `sub foo 1\r\n` are equivalent.

mmmries:
😊  I guess I should have read the documentation before starting to write a client

mmmries:
💥 🎉 💛 💥 🎉 💛 💥 🎉 💛 💥 🎉 💛 💥 🎉 💛 

This is fixed now

PR #24, opened 2017-03-25
Case insensitive commands

tallguy-hackett:
This is my attempt at addressing Issue #18 with case insensitivity while parsing.  I split out some of the message parsing so that the case insensitivity check could work in the `cond` call and renamed a function to more appropriately describe parsing a (NATS) command.

I tried to find some kind of magic voodoo what would allow matching against a regex during a pattern match.  There wasn't anything really good.  The closest I found was using something a lot more complicated than I wanted to get into.  So I settled on using a `cond` to allow for regex match on the command name.

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -96,8 +96,11 @@ defmodule Gnat do
 
   defp perform_handshake(tcp) do
     receive do
-      {:tcp, ^tcp, "INFO"<>_} ->
-        :gen_tcp.send(tcp, "CONNECT {\"verbose\": false}\r\n")
+      {:tcp, ^tcp, operation} ->
+        op_name = operation |> String.split |> List.first
+        if String.match?(op_name, ~r{info}i) do
+          :gen_tcp.send(tcp, "CONNECT {\"verbose\": false}\r\n")
+        end
       after 1000 ->
         {:error, "timed out waiting for info"}
     end
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -135,10 +135,8 @@ defmodule Gnat do
   defp perform_handshake(tcp) do
     receive do
       {:tcp, ^tcp, operation} ->
-        op_name = operation |> String.split |> List.first
-        if String.match?(op_name, ~r{info}i) do
-          :gen_tcp.send(tcp, "CONNECT {\"verbose\": false}\r\n")
-        end
+        "INFO" = operation |> String.split |> List.first |> String.upcase
+        :gen_tcp.send(tcp, "CONNECT {\"verbose\": false}\r\n")
       after 1000 ->
         {:error, "timed out waiting for info"}
     end
--- a/lib/gnat/parser.ex
+++ b/lib/gnat/parser.ex
@@ -13,23 +13,27 @@ defmodule Gnat.Parser do
   end
 
   def parse(parser, "", parsed), do: {parser, Enum.reverse(parsed)}
-  def parse(parser, "PING\r\n", _), do: {parser, [:ping]}
   def parse(parser, bytes, parsed) do
     {index, 2} = :binary.match(bytes, "\r\n")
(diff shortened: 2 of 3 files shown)


tallguy-hackett:
As a note I figure since handling case insensitivity will only matter when handling messages from the server we can send protocol operation names however we desire.  So the operations in question are:

- `INFO`
- `MSG`
- `PING` (sent by both client and server)
- `PONG` (sent by both client and server)
- `+OK` (in verbose mode)
- `-ERR`

I should probably add case insensitivity to handle the `INFO` operation since I just realized it while writing this comment. (whoops!)

mmmries:
hmmmm, the case-insensitive matching is definitely a bit harder to read than the case sensitive version, but supporting does seem important since they've documented it as case-insensitive.

I wonder if we need to support mixed-case versions? `mSg` or `Msg`? If so I wonder if we can write a little macro for case-insensitive pattern matching?

```elixir
parse_case_insensitive("PING") do
  ...
end
```

Could generate multiple function clauses that match on `PING`, `pING`, `PiNG`, etc?

quixoten:
I believe both the Ruby and Go brokers only send UPPERCASE in their implementations. I know the spec states case insensitivity, but it might be more pragmatic to support only the UPPERCASE versions.

mmmries:
Is there a good way to validate that the standard servers all use upcased operations? Or a good way to invalidate that idea? Part of me wants to just implement according to the protocol spec, but if we can reduce the number of variables and be focused only on what we actually need to support that would obviously be nice.

tallguy-hackett:
I feel that it's important to follow the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) and be liberal in what we accept.  Not so much in accepting malformed messages but at least in accepting case insensitive operation names.  (This is probably a result of my networking class in college and the professor beating the principle into our skulls.)  Since the spec dictates that the operation name is case insensitive I feel we should honor that.  Even to the point of accepting mixed case (e.g. `"PiNg"`, `"pinG"`, etc.) operation names.  However, I'm game for whatever the group thinks is best.

film42:
```

**Comment by film42** [comment]
Here's a quick example of gnatsd (go server) accepting mixed case via telnet:
```
# sub here
sUb foo.* 90
+OK

# pub here
pUB foo.bar 5
hello
+OK

# received a message
MSG foo.bar 90 5
hello
```

newellista:
Why don't we just force either upcase or downcase?

`String.downcase(message)`

mmmries:
^^ if we track when we are waiting for the next command we could always read the next 4 bytes, then upcase it and pattern match on that. All commands start with a 3-4 byte opcode.

tallguy-hackett:
@newellista We don't want to downcase the entire message because the contents are case sensitive.  Only the operation name is case insensitive.  I think @mmmries might be on to something with upcasing the operation name.  I wonder if we can upcase it using some piping and then process the operation.  So instead of regex matching we'd insert a call to upcase.  I'll see if I can modify the current code to make it work.

tallguy-hackett:
I was able to take out the regex and pipe the operation name to upcase and then to a new function.  I like this version of the parser a lot more.  Let me know what you guys think. 

I did change the `INFO` as well to just a pattern match after an upcase and rely on a match error if there was a problem.  That doesn't feel like the cleanest way but I wasn't sure if there was a way I like better when using upcase.

It might be worthwhile to add a function clause for an unsupported operation default.  In other words if the server passed something like `"BOGUS\r\n"` currently it would result in an error with no matching clause.

mmmries:
:+1: this has a pretty low-impact on the readability of the code and its runtime performance. I think we can keep this basic structure as we add support for partial messages as well.

I'd be happy to merge this, but let's get feedback from a few other people as well.

newellista:
I like it!

:shipit:

PR #25, opened 2017-03-28
Sub (with support for queue groups)

mmmries:
Adds `SUB` formatting in the `Gnat.Command` module including support for `queue_group`. I also added documentation for subscribing and a test to make sure that we fanout messages in some cases and partition messages when queue group is provided.

/cc @newellista @film42 @tallguy-hackett

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -20,7 +20,7 @@ defmodule Gnat do
 
   def stop(pid), do: GenServer.call(pid, :stop)
 
-  def sub(pid, subscriber, topic), do: GenServer.call(pid, {:sub, subscriber, topic})
+  def sub(pid, subscriber, topic, opts \\ []), do: GenServer.call(pid, {:sub, subscriber, topic, opts})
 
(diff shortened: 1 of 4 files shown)


mmmries:
Fixes #11

lib/gnat.ex
```diff
@@ -20,7 +20,26 @@ defmodule Gnat do
 
   def stop(pid), do: GenServer.call(pid, :stop)
 
-  def sub(pid, subscriber, topic), do: GenServer.call(pid, {:sub, subscriber, topic})
+  @doc """
+  Subscribe to a topic
+
+  By default each subscriber will receive a copy of messages on the topic.
+  When a queue_group is supplied messages will be spread amongs the subscribers
+  int the same group. (see [nats queueing](https://nats.io/documentation/concepts/nats-queueing/))
```
tallguy-hackett: I believe the `int the same` should be `in the same`... no ints here ;)

newellista:
:shipit:

Issue #5, opened 2017-03-23
Handling partial messages in the Parser


mmmries:
We should do this with some good unit tests.

PR #26, opened 2017-03-28
Partial parse

mmmries:
A rough cut at #5

I stole the basic idea here from @newellista who suggested that we only really need to keep the unmatched bytes around for the parser state (no need to have a `state: :waiting` etc). This ends up introducing a few more conditionals which always degrades the readability a bit.

Before I try to refactor I first wanted to get some feedback on whether this looked like a reasonable attempt at parsing.

Benchmarks before and after the change show no significant change in performance (~`81k` parses/sec).

/cc @newellista @tallguy-hackett


newellista:
:shipit:

PR #29, opened 2017-03-28
Add some test helpers

newellista:
In `test_helper.exs`, ensure that `gnatsd` is running before starting
the test run.

Also, add tag that will exclude any tests that have the `@tag: pending`
tag.


mmmries:
:shipit:

PR #35, opened 2017-03-30
Cleanup subscriptions

### LINKED ISSUE #32: Cleanup Subscriptions on unsub
When a client calls `sub` we record a mapping entry to map their `sid` to the receivers address (ie where we will deliver the messages). When a client calls `unsub`, we should remove this mapping. Also if a client calls `unsub` with `max_messages: 1`, we should automatically cleanup the mapping of `sid` => `pid`.

In the `request`/`response` flow we generate a new subscription for every request and failure to cleanup these subscriptions will result in a memory leak.

mmmries:
Fixes #32 

The basic idea here is that we track both a `recipient` pid and a `unsub_after` count for each subscription. If we receive an `unsub` with no max messages we cleanup the subscription immediately after sending the UNSUB command to gnats. If we receive an `unsub` with max messages then we tag the subscription with a number of messages to wait for.

Each time we receive a message and deliver it we decrement the amount remaining and we clean it up when the counter reaches 0.

I tested this by opening `observer` and doing 2MM `Gnat.request` calls. My gnat connection process did not grow in memory. I could see that it's state only ever contained 1 subscription entry and it cleaned up all of the rest of them as responses were received.

/cc @newellista @film42 @jjcarstens @tallguy-hackett

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -23,7 +23,7 @@ defmodule Gnat do
   @doc """
   Subscribe to a topic
 
-  By default each subscriber will receive a copy of messages on the topic.
+  By default each subscriber will receive a copy of every message on the topic.
   When a queue_group is supplied messages will be spread among the subscribers
   in the same group. (see [nats queueing](https://nats.io/documentation/concepts/nats-queueing/))
 
@@ -32,7 +32,7 @@ defmodule Gnat do
 
   ```
   {:ok, gnat} = Gnat.start_link()
-  {:ok, subscription} = Gnat.sub(gnat, "topic")
+  {:ok, subscription} = Gnat.sub(gnat, self(), "topic")
   receive do
     {:msg, %{topic: "topic", body: body}} ->
       IO.puts "Received: \#\{body\}"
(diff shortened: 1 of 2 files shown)


lib/gnat.ex
```diff
@@ -141,10 +135,10 @@ defmodule Gnat do
...
     pub = Command.build(:pub, request.topic, request.body, reply_to: request.inbox)
-    receivers = Map.put(state.receivers, sid, request.recipient)
-    next_sid = sid + 1
     :ok = :gen_tcp.send(state.tcp, [sub, unsub, pub])
-    {:reply, {:ok, sid}, %{state | receivers: receivers, next_sid: next_sid}}
+    state = add_subscription_to_state(state, sid, request.recipient) |> cleanup_subscription_from_state(sid, max_messages: 1)
+    next_sid = sid + 1
+    {:reply, {:ok, sid}, %{state | next_sid: next_sid}}
   end
   def handle_call({:unsub, sid, opts}, _from, state) do
```
film42: While you're here. You should make it so this checks your internal state before sending an unsub command.

mmmries: 👍 in other words don't let the user unsubscribe arbitrary `sid`s?

film42: Exactly 👍 https://github.com/nats-io/pure-ruby-nats/blob/4ac4635c61fdaa132888d254f96b8ea0cf28aaaf/lib/nats/io/client.rb#L352

mmmries: Pushed up the change and added a test to verify it

lib/gnat.ex
```diff
@@ -141,15 +135,19 @@ defmodule Gnat do
...
+    next_sid = sid + 1
+    {:reply, {:ok, sid}, %{state | next_sid: next_sid}}
   end
-  def handle_call({:unsub, sid, opts}, _from, state) do
-    command = Command.build(:unsub, sid, opts)
-    :ok = :gen_tcp.send(state.tcp, command)
-    {:reply, :ok, state}
+  def handle_call({:unsub, sid, opts}, _from, %{receivers: receivers}=state) do
+    case Map.has_key?(receivers, sid) do
+      false -> {:reply, {:error, :no_such_subscription}, state}
```
film42: I would still consider this a success if the caller is cautiously cleaning up a subscription that auto unsub'd, like we do when we perform a double response request. I like that you're distinguishing between what happens, but returning an `error` might be the wrong way to describe it.

film42: Example:
go-nats does not return an error: https://github.com/nats-io/go-nats/blob/master/nats.go#L2147-L2151

pure ruby nats does not return an error: https://github.com/nats-io/pure-ruby-nats/blob/4ac4635c61fdaa132888d254f96b8ea0cf28aaaf/lib/nats/io/client.rb#L352

mmmries: That's a good point. Its not really an error, just it is just a `no-op` essentially. Changed it and removed the test which is pretty much useless.

PR #36, opened 2017-03-31
Auth required

### LINKED ISSUE #27: Support auth_required option when receiving INFO
The NATS protocol for [INFO](http://nats.io/documentation/internals/nats-protocol/#INFO) has an option for `auth_required` indicating that the client should try to authenticate on `CONNECT`.

tallguy-hackett:
This fixes #27 

I added functionality to pass connection information into the parser to generate the `CONNECT` command.  This may not be the best way to do it so I'm very open to suggestions.  I also added a `Gnat.ping` function that sends a `PING` command to the server to help verify an authenticated connection without a need for a subscription request.

There's also no error handling if the client tries to use a connection without authenticating.

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -182,8 +182,7 @@ defmodule Gnat do
   defp perform_handshake(tcp, connection_settings) do
     receive do
       {:tcp, ^tcp, operation} ->
-        {_, [{:info, options}]} = Parser.parse(Parser.new, operation) # |> String.split |> List.first |> String.upcase
-        # :gen_tcp.send(tcp, "CONNECT {\"verbose\": false}\r\n")
+        {_, [{:info, options}]} = Parser.parse(Parser.new, operation)
         connect(tcp, options, connection_settings)
       after 1000 ->
         {:error, "timed out waiting for info"}
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -216,6 +216,7 @@ defmodule Gnat do
   end
   defp process_message(:pong, state) do
     send state.pinger, :pong
+    state
   end
 
   defp update_subscriptions_after_delivering_message(%{receivers: receivers}=state, sid) do
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -95,9 +95,10 @@ defmodule Gnat do
   Ping the NATS server
 
   This correlates to the [PING](http://nats.io/documentation/internals/nats-protocol/#PINGPONG) command in the NATS protocol.
+  If the NATS server responds with a PONG message this function will return `:ok`
   ```
   {:ok, gnat} = Gnat.start_link()
-  {:pong} = Gnat.ping(gnat)
+  :ok = Gnat.ping(gnat)
   ```
   """
   def ping(pid) do
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -5,6 +5,7 @@
 defmodule Gnat do
   use GenServer
   require Logger
+  require Poison
   alias Gnat.{Command, Parser}
 
   @default_connection_settings %{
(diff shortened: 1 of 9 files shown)


tallguy-hackett:
This probably fails because for the test I wrote there needs to be a second server running that requires authentication.

mmmries:
@tallguy-hackett you could get travis to pass by adding a `before_script` step that starts an additional gnats server on a different port

mmmries:
Looks like you and @film42 have both been thinking about this problem: https://github.com/mmmries/gnat/compare/gt/parse_info_and_ok?expand=1

I like that both of you pushed the parsing into `Gnat.Parser`.

lib/gnat.ex
```diff
@@ -90,10 +91,29 @@ defmodule Gnat do
...
+  Ping the NATS server
+
+  This correlates to the [PING](http://nats.io/documentation/internals/nats-protocol/#PINGPONG) command in the NATS protocol.
+  If the NATS server responds with a PONG message this function will return `:ok`
+  ```
+  {:ok, gnat} = Gnat.start_link()
+  :ok = Gnat.ping(gnat)
+  ```
+  """
+  def ping(pid) do
```
mmmries: Exposing this as usable behavior seems odd to me. What happens for instance if two different clients both send call `Gnat.ping` close to the same time? The `state.pinger` would end up holding the address of the second client and the first one would timeout waiting for a `PONG`.

I think it probably makes sense for the `Gnat` actor to send a `PING` every so often and expect a response, but I don't understand the purpose of making it a callable behavior. Maybe I'm missing something?

test/gnat_test.exs
```diff
@@ -8,6 +8,20 @@ defmodule GnatTest do
...
+    connection_settings = %{
+      host: 'localhost',
+      port: 4223,
+      tcp_opts: [:binary],
+      username: "bob",
+      password: "alice"
+    }
+    {:ok, pid} = Gnat.start_link(connection_settings)
+    assert Process.alive?(pid)
+    :ok = Gnat.ping(pid)
```
mmmries: oh, is this the reason you wanted expose `ping` as a public function? I guess it's a pretty good way to verify that the connection was successful...  It's certainly simpler than doing a `sub` + `pub` + `receive`

test/test_helper.exs
```diff
@@ -11,3 +11,16 @@ case :gen_tcp.connect('localhost', 4222, [:binary]) do
               " #{:inet.format_error(reason)}\n" <>
               "You probably need to start gnatsd."
 end
+
+case :gen_tcp.connect('localhost', 4223, [:binary]) do
+  {:ok, socket} ->
+    :gen_tcp.close(socket)
+  {:error, reason} ->
+    Mix.raise "Cannot connect to gnatsd" <>
+              " (http://localhost:4223):" <>
```
mmmries: I like the helpful error message here, but it seems like a pain to force all developers to run two gnatsd servers while developing. Maybe this is a good time to tag our tests and default the integration tests to be skipped

lib/gnat.ex
film42: This is actually pretty common on nats clients, but the method is called `flush`: https://github.com/nats-io/go-nats/blob/master/nats.go#L2464-L2468

lib/gnat.ex
```diff
@@ -149,17 +169,29 @@ defmodule Gnat do
...
+    {:reply, :ok, Map.put(state, :pinger, pinger)}
+  end
 
-  defp perform_handshake(tcp) do
+  defp perform_handshake(tcp, connection_settings) do
     receive do
       {:tcp, ^tcp, operation} ->
-        "INFO" = operation |> String.split |> List.first |> String.upcase
-        :gen_tcp.send(tcp, "CONNECT {\"verbose\": false}\r\n")
+        {_, [{:info, options}]} = Parser.parse(Parser.new, operation)
```
film42: As far as I know, there's no guarantee that the entire INFO message will fit in this first TCP packet, so I think moving the message parsing into Parser is a good start, but I think it might be better to let the genserver receive packets until the parser finds the INFO message. And then maybe we wait for a message from the genserver in `init` to ensure the handshake completed successfully?

lib/gnat/parser.ex
```diff
@@ -1,5 +1,6 @@
 defmodule Gnat.Parser do
   require Logger
+  require Poison
```
film42: I don't think we need to require poison to use the parser.

tallguy-hackett:
I added some functionality for tagging tests as well as dynamically checking for a second server when including the `multi_server` tagged tests.  I think it works out pretty well.  So now by default running tests won't include the requirement for the 2nd server.

lib/gnat.ex
```diff
@@ -149,17 +168,29 @@ defmodule Gnat do
...
+        connect(tcp, options, connection_settings)
       after 1000 ->
         {:error, "timed out waiting for info"}
     end
   end
 
+  defp connect(tcp, %{auth_required: true}=_options, %{username: username, password: password}=_connection_settings) do
```
jjcarstens: I don't think we need to assign unused variables `_connection_settings` and `_options` here. The pattern matching should be just fine and if someone wanted context about _what_ is being matched, they can look and the other `connect` function

PR #37, opened 2017-04-13
Basic tls

### LINKED ISSUE #30: Support TLS Connections

mmmries:
Resolves #30 

This PR just handles the very basic case where the client has been started with `tls: true` and the server has `tls_required: true` settings. I will follow it up with at least 2 more PRs:
* handle cases where the server and client mismatch on requiring TLS
* refactoring the handshake into its own module and passing the entire state to `socket_write` and `socket_close` rather than pulling pieces out of the state

I wanted to get this PR opened before tackling either of those other changes since it is already a bit large.

/cc @tallguy-hackett @newellista @film42 @jjcarstens


mmmries:
💥 Travis ran for 44 seconds to do all test cases

PR #38, opened 2017-04-13
Handshake refactor

mmmries:
While working on #37 I was bugged by the fact that we had several functions in the `Gnat` module that were only useful for the handshake. I thought it would be nicer to let all of the handshake logic happen in its own module with just a simple `{:ok, socket}` or `{:error, reason}` return value.

I also didn't like constantly having to to do

```elixir
socket_write(state.connection_settings, state.socket, "PING\r\n")
```

So I changed the `socket_write` and `socket_close` functions to just accept the entire state as the first argument and it extracts the options/socket as needed.

```elixir
socket_write(state, "PING\r\n")
```

/cc @quixoten @film42 @newellista @jjcarstens @tallguy-hackett

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -139,25 +139,25 @@ defmodule Gnat do
 
 
   def handle_call(:stop, _from, state) do
-    socket_close(state.connection_settings, state.socket)
+    socket_close(state)
     {:stop, :normal, :ok, state}
   end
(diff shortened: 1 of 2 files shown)


quixoten:
:+1:

film42:
I'm not sure what kind of guarantees socket write has, but it might be better to queue these messages up in the event that your kernel's buffer fills.

PR #39, opened 2017-04-14
Client cert

mmmries:
💥 💥 💥 💥 💥 💥 

I wrote a failing test assuming that I would have to fix it with implementation code, but since we already allow passing in of the `ssl_opts`, this turned out to just be a matter of passing in connection options.

I went ahead and documented some of the connection options and this documentation looks like this when rendered:
screen shot 2017-04-13 at 9 19 13 pm

/cc @film42 @newellista @jjcarstens @quixoten @tallguy-hackett


quixoten:
Nice! :+1:

PR #47, opened 2017-04-20
GenServer opts

mmmries:
allow gen server options to be passed to `Gnat` for things like registered names. This is useful for having a supervised nats connection with a registered name. I'm using this in my application at work so that I can pass the registered name as the first argument to `Gnat.sub`, `Gnat.pub`, etc.

Also updated the documentation.

/cc @jjcarstens @newellista @film42 @tallguy-hackett

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -15,8 +15,6 @@ defmodule Gnat do
     tls: false,
   }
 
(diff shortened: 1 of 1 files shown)


newellista:
:shipit:

Issue #42, opened 2017-04-20
Save broker INFO in the Gnat state so it's available if the connection crashes


mmmries:

mmmries:
This will be part of #54

PR #48, opened 2017-04-21
Handle various tcp error

mtokioka:
Add some error handling.
GenServer should be stopped when server is down.

this may conflict the concept of #42. But I like simple client;)


mmmries:
Woohoo! Thanks @mtokioka! I just noticed the issue with `{:tcp_closed...` issue yesterday when I was testing what happened when the broker was shutdown.

PR #50, opened 2017-04-29
Support configurable connection timeout

quixoten:
This adds support for a configurable connection timeout that defaults to
3 seconds. I also changed the return value in the init function under
error conditions to `{:stop, reason}` from `{:error, reason}`. I think
this is correct because `{:error, reason}` generated:

```
** (EXIT from #PID<0.208.0>) bad return value: {:error, :timeout}
```

I could also not find support for `{:error, _}` in the docs:
https://hexdocs.pm/elixir/GenServer.html#c:init/1

This was tricky to test because calling `Gnat.start_link` killed the
test immediately. Hopefully calling `Gnat.init` instead is a viable way
to test the behavior.

---

@mmmries @film42 @newellista @tallguy-hackett @jjcarstens

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -11,7 +11,7 @@ defmodule Gnat do
     host: 'localhost',
     port: 4222,
     tcp_opts: [:binary],
-    timeout: 3_000,
+    connection_timeout: 3_000,
     ssl_opts: [],
     tls: false,
   }
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -11,6 +11,7 @@ defmodule Gnat do
     host: 'localhost',
     port: 4222,
     tcp_opts: [:binary],
+    timeout: 3_000,
     ssl_opts: [],
     tls: false,
   }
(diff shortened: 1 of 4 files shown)


lib/gnat/handshake.ex
```diff
@@ -5,8 +5,10 @@ defmodule Gnat.Handshake do
   This module provides a single function which handles all of the variations of establishing a connection to a gnatsd server and just returns {:ok, socket} or {:error, reason}
   """
   def connect(settings) do
-    {:ok, tcp} = :gen_tcp.connect(settings.host, settings.port, settings.tcp_opts)
-    perform_handshake(tcp, settings)
+    case result = :gen_tcp.connect(settings.host, settings.port, settings.tcp_opts, settings.timeout) do
```
jjcarstens: We shouldn't need the `result =` assignment here is the case condition. It can be assigned on matching below

```elixir
{:ok, tcp} -> perform_handsake(tcp, settings)
result -> result
...
```

lib/gnat.ex
```diff
@@ -11,6 +11,7 @@ defmodule Gnat do
     host: 'localhost',
     port: 4222,
     tcp_opts: [:binary],
+    timeout: 3_000,
```
mmmries: should we call this a `connect_timeout`? Looking at it again today I think the name `timeout` might confusing

quixoten:
Updated.

mmmries:
The failed builds last night were caused by failed `wget` when travis was trying to install elixir. It was not related to the project.

PR #56, opened 2017-05-30
cleanup state after unsub

mtokioka:
subscription information is not deleted when calling ```Gnat.unsub(state, sid)``` (without max_messages).


mmmries:
:shipit: thanks @mtokioka looks like we spacd on the manual unsub case.

mmmries:
Released as `0.3.2` https://hex.pm/packages/gnat/0.3.2

PR #59, opened 2017-07-02
Supervised consumers

mmmries:
This is a first cut at having supervised consumers. This feature can be used to support things like RPC servers where a client application wants to have long-lived subscriptions that last across re-connect attempts. It works by adding an entry to your supervision tree like:

```elixir
consumer_settings = %{
  connection_name: :rpc_gnat,
  consuming_function: {Dispatch.RpcServer, :handle_request},
  subscription_topics: [
    %{topic: "rpc.atlas.dispatch.shard_service.search", queue_group: "rpc.atlas.dispatch.shard_service.search"},
    %{topic: "rpc.atlas.dispatch.user_shard_service.search", queue_group: "rpc.atlas.dispatch.user_shard_service.search"},
  ],
}

children = [worker(Gnat.ConsumerSupervisor, [settings, [name: :rpc_consumer]], shutdown: 30_000)]
supervise(children, strategy: :one_for_one)
```

This consumer supervisor will check for the named connection, and if it finds the connection available it will subscribe to the given topics and monitor that connection. It automatically retries in the case that the connection dies or is unavailable. It will then call the module and function you specify whenever a message arrives for one of those subscriptions.

In a future cut I will add graceful shutdown that unsubscribes itself from the topics and waits for all of the outstanding tasks to complete before finishing its shutdown. How long to wait can be configured in the supervision tree via the `:shutdown` option which will give the `ConsumerSupervisor` a configurable number of milliseconds to finish its own shutdown before it is sent a hard kill signal and forced to shutdown.

I also need to add a moduledoc and perhaps a few sanity tests in a followup PR

/cc @newellista @film42 @abrandoned

### DIFF
--- a/lib/gnat/consumer_supervisor.ex
+++ b/lib/gnat/consumer_supervisor.ex
@@ -1,0 +1,58 @@
+defmodule Gnat.ConsumerSupervisor do
+  use GenServer
+  require Logger
+
+  def start_link(settings, options) do
+    GenServer.start_link(__MODULE__, settings, options)
+  end
+
+  def init(settings) do
+    Process.flag(:trap_exit, true)
+    {:ok, task_supervisor_pid} = Task.Supervisor.start_link()
+    connection_name = Map.get(settings, :connection_name)
+    subscription_topics = Map.get(settings, :subscription_topics)
+    consuming_function = Map.get(settings, :consuming_function)
+    send self(), :connect
+    state = %{
+      connection_name: connection_name,
+      connection_pid: nil,
+      consuming_function: consuming_function,
+      status: :disconnected,
+      subscription_topics: subscription_topics,
+      subscriptions: [],
+      task_supervisor_pid: task_supervisor_pid,
+    }
+    {:ok, state}
+  end
+
+  def handle_info(:connect, %{connection_name: name}=state) do
+    case Process.whereis(name) do
+      nil ->
+        Process.send_after(self(), :connect, 2_000)
+        {:noreply, state}
+      connection_pid ->
+        _ref = Process.monitor(connection_pid)
+        subscriptions = Enum.map(state.subscription_topics, fn(topic_and_queue_group) ->
+          topic = Map.fetch!(topic_and_queue_group, :topic)
+          queue_group = Map.get(topic_and_queue_group, :queue_group)
+          {:ok, subscription} = Gnat.sub(connection_pid, self(), topic, queue_group: queue_group)
+          subscription
+        end)
+        {:noreply, %{state | status: :connected, connection_pid: connection_pid, subscriptions: subscriptions}}
+    end
+  end
+
+  def handle_info({:DOWN, _ref, :process, connection_pid, _reason}, %{connnection_pid: connection_pid}=state) do
+    Process.send_after(self(), :connect, 2_000)
+    {:noreply, %{state | status: :disconnected, connection_pid: nil, subscriptions: []}}
+  end
+
+  def handle_info({:msg, gnat_message}, %{consuming_function: {mod, fun}}=state) do
+    Task.Supervisor.async_nolink(state.task_supervisor_pid, mod, fun, [gnat_message])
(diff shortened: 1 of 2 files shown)


lib/gnat/consumer_supervisor.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  def handle_info(:connect, %{connection_name: name}=state) do
+    case Process.whereis(name) do
+      nil ->
+        Process.send_after(self(), :connect, 2_000)
+        {:noreply, state}
+      connection_pid ->
+        _ref = Process.monitor(connection_pid)
+        subscriptions = Enum.map(state.subscription_topics, fn(topic_and_queue_group) ->
+          topic = Map.fetch!(topic_and_queue_group, :topic)
```
film42: Will this cause us to respawn the genserver over and over again if this blows, or is there an exit strategy?

lib/gnat/consumer_supervisor.ex
```diff
@@ -0,0 +1,66 @@
...
+    connection_name = Map.get(settings, :connection_name)
+    subscription_topics = Map.get(settings, :subscription_topics)
+    consuming_function = Map.get(settings, :consuming_function)
+    send self(), :connect
+    state = %{
+      connection_name: connection_name,
+      connection_pid: nil,
+      consuming_function: consuming_function,
+      status: :disconnected,
+      subscription_topics: subscription_topics,
```
film42: What about sub topics that are made after the supervisor is created? Does this list get stale?

lib/gnat/consumer_supervisor.ex
mmmries: OTP supervisors will bail on restarting a child process if it is being restarted too frequently. By default the limit is 3 restarts in 5 seconds will cause the supervisor to fail and will prevent the application from completing the startup process. [These can be tuned here](https://hexdocs.pm/elixir/Supervisor.Spec.html#supervise/2-options)

lib/gnat/consumer_supervisor.ex
mmmries: Sub-topics or transient topics can still be handled by the gnat connection directly (ie setting up temporary inbox topics for replies etc), but this PR is specifically aimed at subscriptions that are being maintained for the lifetime of an application.

In `protobuf-nats` I thought we generated a fixed set of topics at startup and those remained constant? Do we have transient topics for anything other than RPC requests?

film42: You're correct, but it feels a bit weird to differentiate topics that last for the lifetime of the process vs connection. IMO all sub topics should be treated the same.

mmmries: I think I'm misunderstanding what you mean by sub topics. As far as the `Gnat` connection is concerned all topics and subscriptions are the exact same. This consumer code is just meant to be some helpers around automatically subscribing to topics that you want to keep your application listening for. So when the connection comes and goes you automatically re-create the long-lived topics.

Are you thinking that the connection should remember previous subscriptions and automatically re-subscribe to them when a new connection is formed? That would be part of the responsibility of the `Gnat` connection rather than a separate responsibility?

mmmries:
I did some benchmarking on 8-core nodes and adding in the supervised tasks adds a little overhead. A previous benchmark that setup 4 connections and spawned the handling of those messages as raw unlinked processes was able to complete `10,258` request/sec. A similar version with 4 supervised connections and 4 supervised consumers did `9,595` request/sec (`94.5`%).

I think that 5.5% is worth giving up in order to get SASL error reports when the processing of a message fails and have graceful shutdown semantics.

mmmries:
I'm going to go ahead and release this change, but document it as experimental. I'm working an a potential alternative where people send messages to a managing process that automatically re-subscribes when a connection is re-established. That way consumers don't have to monitor the connection or worry about whether it currently exists. I want to experiment both types of solution so I can better learn the pros and cons.

PR #61, opened 2017-07-07
accept multiple pub commands and send them in a single TCP command

mmmries:
I always wondered how much of a bottleneck it was for us to write small publishing packets to the TCP/SSL socket for each command so I decided to use the `pub128` benchmark in this project and experiment with the idea of accepting multiple `pub` commands at the same time, combining them into a single TCP send and then doing a `GenServer.reply` for each one of them.

This change almost doubled the publishing throughput. Below you can see the throughput numbers for various level of publishing clients and with/without tcp nodelay on the connection.

gnat_publish_optimize

> Using a localhost connection in this case was an attempt to narrow this down to CPU overhead and/or erlang scheduling overhead (ignoring networking performance for the moment). But maybe I should repeat these tests with a multi-node setup so we can test over an actual ethernet connection?

I'm curious if we think this type of optimization is worth it. The code is a bit more complex, but a nearly 2x improvement could be useful in some cases.

/cc @film42 @quixoten @abrandoned @newellista

### DIFF
--- a/bench.exs
+++ b/bench.exs
@@ -28,19 +28,19 @@ msg16="74c93e71c5aa03ad"
 
 tcp_packet = "MSG topic 1 128\r\n#{msg128}\r\n"
 Benchee.run(%{
-  "parse-128" => fn -> {_parser, [_msg]} = Gnat.Parser.new() |> Gnat.Parser.parse(tcp_packet) end,
+ "parse-128" => fn -> {_parser, [_msg]} = Gnat.Parser.new() |> Gnat.Parser.parse(tcp_packet) end,
   "pub - 128" => fn -> :ok = Gnat.pub(pid, "pub128", msg128) end,
-  "sub-unsub-pub-16" => fn ->
-    rand = :crypto.strong_rand_bytes(8) |> Base.encode64
-    {:ok, subscription} = Gnat.sub(pid, self(), rand)
-    :ok = Gnat.unsub(pid, subscription, max_messages: 1)
-    :ok = Gnat.pub(pid, rand, msg16)
-    receive do
-      {:msg, %{topic: ^rand, body: ^msg16}} -> :ok
-      after 100 -> raise "timed out on sub"
-    end
-  end,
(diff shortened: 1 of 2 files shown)


quixoten:
I don't know that there's any urgency to push past the current performance numbers. I can't think of a scenario within our usage it would be a bottleneck at the current rate. There's a lot of time to experiment if you think a better implementation might manifest.

I found similar results in ruby. Even with a large SND_BUF on the socket, it was always faster to buffer outside the socket first. If memory serves, the `C`, `Ruby`, and `Go` clients all use a separate write thread to flush a buffer across the socket in a loop. I was hoping Erlang might do something similar internally. Is it possible to tweak buffering on the gentcp socket and effect throughput? What would it look like in Erlang to have a separate writer process?

mmmries:
I definitely agree that this is not a necessary or urgent optimization. I'm on vacation, so I get to write whatever code sounds interesting 😎 

> C, Ruby, and Go clients all use a separate write thread to flush a buffer across the socket in a loop.  I was hoping Erlang might do something similar internally.

Currently we do the reading/parsing/dispatching all int the same actor that is doing the formatting and writing of commands to the broker. To the best of my knowledge the `gen_tcp` process that we open is actually a c-extension that just wraps system calls and behaves like an actor.

>  Is it possible to tweak buffering on the gentcp socket and effect throughput?

Yes, when you open a `gen_tcp` process you can give it settings like a write buffer size etc.

> What would it look like in Erlang to have a separate writer process?

I've been thinking about this question and I think there are a couple of ways to do it, but I'll start with the one that seems most obvious to me where we actually spawn 2 processes when starting a `Gnat` connection, the second process is just for writing to the socket and the first will receive bytes from the connection and do the parsing/dispatching.

newellista:
I don't find the code that much more complex, and a 2x speedup is always welcome.  While there may be a better way to do this, and we don't _need_ the performance right now, I believe this is a worthwhile change.

:shipit:

mmmries:
Closing in favor of #65

PR #65, opened 2017-10-10
Optimize Throughput

mmmries:
This follows the pattern in #61, but follows the same pattern for accepting data from the gnats brokers. This gives us ~50% increase in throughput on a [realistic benchmark](https://gist.github.com/mmmries/cf86af5149524155fc2eac3df4a9d948) that sends 128byte requests and gets back 1.3kb responses.

I've also added a `:gnat` key to the messages that get delivered so that client applications can easily setup multiple connections and send back a reply based on the data in the message they received. (ie they don't need to track the name or pid of a connection)

/cc @newellista @film42


jjcarstens:
This is awesome... 🎉

PR #66, opened 2017-10-11
Run mix format mix.exs "lib/**/*.{ex,exs}" "test/**/*.{ex,exs}"

film42:
Nothing else but mix format.

cc @mmmries @jjcarstens @newellista

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -7,15 +7,15 @@ defmodule Gnat do
   require Logger
   alias Gnat.{Command, Parser}
 
-  @type message :: %{topic: String.t, body: String.t, reply_to: String.t}
+  @type message :: %{topic: String.t(), body: String.t(), reply_to: String.t()}
 
   @default_connection_settings %{
     host: 'localhost',
     port: 4222,
     tcp_opts: [:binary],
-    connection_timeout: 3_000,
+    connection_timeout: 3000,
     ssl_opts: [],
-    tls: false,
+    tls: false
   }
 
   @doc """
@@ -34,7 +34,7 @@ defmodule Gnat do
 
   The final `opts` argument will be passed to the `GenServer.start_link` call so you can pass things like `[name: :gnat_connection]`.
   """
-  @spec start_link(map(), keyword()) :: GenServer.on_start
+  @spec start_link(map(), keyword()) :: GenServer.on_start()
   def start_link(connection_settings \\ %{}, opts \\ []) do
     GenServer.start_link(__MODULE__, connection_settings, opts)
   end
@@ -47,7 +47,7 @@ defmodule Gnat do
   :ok = Gnat.stop(gnat)
   ```
   """
-  @spec stop(GenServer.server) :: :ok
+  @spec stop(GenServer.server()) :: :ok
   def stop(pid), do: GenServer.call(pid, :stop)
 
   @doc """
@@ -69,8 +69,9 @@ defmodule Gnat do
   end
   ```
   """
-  @spec sub(GenServer.server, pid(), String.t, keyword()) :: {:ok, non_neg_integer()}
-  def sub(pid, subscriber, topic, opts \\ []), do: GenServer.call(pid, {:sub, subscriber, topic, opts})
+  @spec sub(GenServer.server(), pid(), String.t(), keyword()) :: {:ok, non_neg_integer()}
+  def sub(pid, subscriber, topic, opts \\ []),
+    do: GenServer.call(pid, {:sub, subscriber, topic, opts})
 
   @doc """
   Publish a message
@@ -88,7 +89,7 @@ defmodule Gnat do
   :ok = Gnat.pub(gnat, "characters", "Star Lord", reply_to: "me")
   ```
   """
-  @spec pub(GenServer.server, String.t, binary(), keyword()) :: :ok
+  @spec pub(GenServer.server(), String.t(), binary(), keyword()) :: :ok
   def pub(pid, topic, message, opts \\ []), do: GenServer.call(pid, {:pub, topic, message, opts})
 
   @doc """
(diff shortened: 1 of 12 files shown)


lib/gnat.ex
```diff
@@ -69,8 +69,9 @@ defmodule Gnat do
   end
   ```
   """
-  @spec sub(GenServer.server, pid(), String.t, keyword()) :: {:ok, non_neg_integer()}
-  def sub(pid, subscriber, topic, opts \\ []), do: GenServer.call(pid, {:sub, subscriber, topic, opts})
+  @spec sub(GenServer.server(), pid(), String.t(), keyword()) :: {:ok, non_neg_integer()}
+  def sub(pid, subscriber, topic, opts \\ []),
+    do: GenServer.call(pid, {:sub, subscriber, topic, opts})
```
jjcarstens: I'm not a huge fan of changes like this which is making a newline if line > 80 characters.

I'd say if its close, leave as a single line. Otherwise lets block it like
```elixir
def sub(pid, subscriber, topic, opts \\ [])
  GenServer.call(pid, {:sub, subscriber, topic, opts})
end

File: lib/gnat.ex [hunk 2]
```diff
@@ -108,14 +109,19 @@ defmodule Gnat do
...
+          {:ok, message} | {:error, :timeout}
   def request(pid, topic, body, opts \\ []) do
-    receive_timeout = Keyword.get(opts, :receive_timeout, 60_000)
-    inbox = "INBOX-#{:crypto.strong_rand_bytes(12) |> Base.encode64}"
-    {:ok, subscription} = GenServer.call(pid, {:request, %{recipient: self(), inbox: inbox, body: body, topic: topic}})
+    receive_timeout = Keyword.get(opts, :receive_timeout, 60000)
+    inbox = "INBOX-#{:crypto.strong_rand_bytes(12) |> Base.encode64()}"
+
+    {:ok, subscription} =
+      GenServer.call(pid, {:request, %{recipient: self(), inbox: inbox, body: body, topic: topic}})
```
jjcarstens: I think we could benefit here by just pulling the map into its own variable:

```elixir
...
request = %{recipient: self(), inbox: inbox, body: body, topic: topic}
{:ok, subscription} = GenServer.call(pid, {:request, request})
```

lib/gnat.ex
```diff
@@ -152,49 +158,70 @@ defmodule Gnat do
...
+        {
+          :ok,
+          %{
+            socket: socket,
+            connection_settings: connection_settings,
+            next_sid: 1,
+            receivers: %{},
+            parser: parser
+          }
+        }
```
jjcarstens: Similar here. Formatter is trying to avoid long lines, but so much space in this tuple is odd to me. Maybe instead?
```elixir
response = %{socket: socket, connection_settings: connection_settings, next_sid: 1, receivers: %{}, parser: parser}
{:ok, response}
```

lib/gnat.ex
```diff
@@ -247,48 +285,59 @@ defmodule Gnat do
...
     %{state | receivers: receivers}
   end
 
   defp process_message({:msg, topic, sid, reply_to, body}, state) do
     unless is_nil(state.receivers[sid]) do
-      send state.receivers[sid].recipient, {:msg, %{topic: topic, body: body, reply_to: reply_to, gnat: self()}}
+      send(state.receivers[sid].recipient, {
+        :msg,
+        %{topic: topic, body: body, reply_to: reply_to, gnat: self()}
+      })
```
jjcarstens: Similar here. Splitting out the map to make the tuple on the same line would be 👌

lib/gnat/command.ex
```diff
@@ -4,12 +4,27 @@ defmodule Gnat.Command do
...
+      @pub,
+      " ",
+      topic,
+      " ",
+      reply,
+      " #{IO.iodata_length(payload)}",
+      @newline,
+      payload,
+      @newline
+    ]
```
jjcarstens: We probably should put this in a function block rather than shorthand

lib/gnat/consumer_supervisor.ex
```diff
@@ -39,86 +39,114 @@ defmodule Gnat.ConsumerSupervisor do
...
+
+        {
+          :noreply,
+          %{
+            state
+            | status: :connected,
+              connection_pid: connection_pid,
+              subscriptions: subscriptions
+          }
+        }
```
jjcarstens: This is kind of worse..

lib/gnat/handshake.ex
```diff
@@ -15,29 +16,38 @@ defmodule Gnat.Handshake do
...
   defp socket_write(%{tls: true}, socket, iodata), do: :ssl.send(socket, iodata)
   defp socket_write(_, socket, iodata), do: :gen_tcp.send(socket, iodata)
 
-  defp send_connect_message(socket, %{auth_required: true}=_options, %{username: username, password: password}=connection_settings) do
-    opts = Poison.Encoder.encode(%{user: username, pass: password, verbose: false}, strict_keys: true)
+  defp send_connect_message(
+         socket,
+         %{auth_required: true} = _options,
+         %{username: username, password: password} = connection_settings
+       ) do
```
jjcarstens: We probably should just remove `_options` since it isn't even used

film42:
So here are the options. We should probably just embrace the change though:


## Options
* `:file` - the file which contains the string, used for error
reporting
* `:line` - the line the string starts, used for error reporting
* `:line_length` - the line length to aim for when formatting
the document. Defaults to 98.
* `:locals_without_parens` - a keyword list of name and arity
pairs that should be kept without parens whenever possible.
The arity may be the atom `:*`, which implies all arities of
that name. The formatter already includes a list of functions
and this option augments this list.
* `:rename_deprecated_at` - rename all known deprecated functions
at the given version to their non-deprecated equivalent. It
expects a valid `Version` which is usually the minimum Elixir
version supported by the project.

lib/gnat.ex
```diff
@@ -7,15 +7,15 @@ defmodule Gnat do
...
 
   @default_connection_settings %{
     host: 'localhost',
     port: 4222,
     tcp_opts: [:binary],
-    connection_timeout: 3_000,
+    connection_timeout: 3000,
     ssl_opts: [],
-    tls: false,
+    tls: false
```
mmmries: no more trailing commas 😭

lib/gnat.ex
newellista: I actually like this format rule.  I'm willing to trade the extra whitespace to make the map significantly more readable.

mmmries:
This PR has fallen a bit behind. We ended up going with a different strategy where we are going to slowly add patterns to the `.formatter.exs` file so we can make smaller decisions about what to format and how to comply with the formatter as we go. See #84 for the beginnings of this effort

PR #69, opened 2017-11-10
Use a shared wildcard subscription to process all inbox messages

film42:
It turns out subscribing and unsubscribing frequently is not good for NATS. It needs to lock its subscription list before it can perform an update, so churning through subscriptions will slow your entire system down. The good people at apcera started migrating to this pattern. This is my first attempt at making it work. Here's what's changed:

1. When we call `Gnat.start_link` we create a new subscription with SID `0` to the `_INBOX.<random_id_for_the_connection>.*` topic. Example: ` _INBOX.Jhf7AcTGP3x4dAV9.*`.
2. There's a new `Gnat.new_inbox` method that generates a new random inbox appended to the connection inbox prefix. Example: `_INBOX.Jhf7AcTGP3x4dAV9.gV4Xzwd0BmdmJMBx`.
3. The `Gnat.sub` method now has an `as_request: true` option which can be provided to update the request receivers map instead of creating a new subscription. When `as_request: true` is provided, we return the inbox string as the SID. This let's us know _which_ key we should remove from the request receivers map when we unsub. If the topic provided to `Gant.sub` with `as_request: true` doesn't contain the connection inbox prefix, we return an error. I don't love this. Any feedback? 
4. The `Gnat.unsub` method can now looks to see if the SID is a binary or number. If a binary, we attempt to remove it from the new request receivers map. Otherwise, same old story. The only weird thing here is that there's no support for `max_messages: N` with a "request subscription". I'm open to adding this here or in a follow up PR. Thoughts?
5. Finally, the `Gnat.request` method functions the same. Only the insides have been altered to be more efficient. There's still room for improvement (like removing the explicit `unsubscribe`).

cc @mmmries @newellista @jjcarstens

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -18,6 +18,8 @@ defmodule Gnat do
     tls: false,
   }
 
+  @request_sid 0
+
   @doc """
   Starts a connection to a nats broker
 
@@ -273,9 +275,9 @@ defmodule Gnat do
   defp create_request_subscription(%{request_inbox_prefix: request_inbox_prefix}=state) do
     # Example: "_INBOX.Jhf7AcTGP3x4dAV9.*"
     wildcard_inbox_topic = request_inbox_prefix <> "*"
-    sub = Command.build(:sub, wildcard_inbox_topic, 0, [])
+    sub = Command.build(:sub, wildcard_inbox_topic, @request_sid, [])
     :ok = socket_write(state, [sub])
-    add_subscription_to_state(state, 0, self())
+    add_subscription_to_state(state, @request_sid, self())
   end
 
   defp nuid(), do: :crypto.strong_rand_bytes(12) |> Base.encode64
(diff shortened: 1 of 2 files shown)


lib/gnat.ex
```diff
@@ -263,6 +343,15 @@ defmodule Gnat do
     %{state | receivers: receivers}
   end
 
+  defp process_message({:msg, topic, @request_sid, reply_to, body}, state) do
+    if Map.has_key?(state.request_receivers, topic) do
+      send state.request_receivers[topic], {:msg, %{topic: topic, body: body, reply_to: reply_to, gnat: self()}}
+      state
+    else
+      Logger.error "#{__MODULE__} got a response for a request, but that is no longer registered"
```
film42: Is this going to be annoying?

mmmries: I think we should probably report an error via logger and via the `:error_logger` http://erlang.org/doc/man/error_logger.html. It's part of the standard library and tools like the honeybadger reporter hook into that.

film42: @mmmries I wonder if we should just delete this line. If we time out a request and unsub but then later receive that message (ex: db was paused), it will his this code path. Probably doesn't warrant a honey badger style message, or maybe even a log line. If that's the case then we can delete it.

mmmries: Maybe we should just log it with the `debug` level? that way if someone's trying to figure out where their messages are going they can see that timed out requests are showing up later?

jjcarstens: I would vote `debug` log. That way it definitely can help identify a possible issue if the developer is trying to look for it.

mmmries:
This was superceded by #73

PR #71, opened 2018-05-06
add support for parsing +OK messages and property tests

mmmries:
Adds another 2 property tests for the parser that will randomly generate all kinds of protocol messages (`PING`,`PONG`,`+OK`,`-ERR...` and `MSG...`). One test feeds these protocol messages into the parser one at a time, the other test glues them together into one big string and then randomly chunks them up to make sure that we can parse them with any arbitrary byte splitting that might happen at the networking level.

### DIFF
--- a/test/support/generators.ex
+++ b/test/support/generators.ex
@@ -5,9 +5,25 @@ defmodule Gnat.Generators do
     elements('0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ')
   end
 
-  # subscription names (subject names with * and > added)
   def delimiter, do: let(chunks <- non_empty(list(delimiter_char())), do: Enum.join(chunks, ""))
+
   def delimiter_char, do: union([" ","\t"])
+
+  def error do
+    let chars <- list(error_char) do
+      %{binary: "-ERR '#{List.to_string(chars)}'\r\n"}
+    end
+  end
+
+  def error_char do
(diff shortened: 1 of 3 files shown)


test/support/generators.ex
```diff
@@ -46,6 +62,10 @@ defmodule Gnat.Generators do
 
   def payload(size), do: binary(size)
 
+  def protocol_message do
+    union([ok(), ping(), pong(), error(), message()])
```
film42: I think INFO might be good to add here.

mmmries: :+1: you are totally right, just forgot about it, but we definitely parse it. Will add

mmmries:
Out of curiosity I tried running the property tests for a longer period with the following command:

```
mmmries-pro:gnat$ N=100000 mix test --only property --timeout 5184000
Compiling 1 file (.ex)
Including tags: [:property]
Excluding tags: [:test, :pending, :multi_server]

....

Finished in 4394.9 seconds
4 properties, 32 tests, 0 failures, 32 skipped
```

Setting `N` to 100k means that we will generate 1MM random message binaries and 300k other test cases for the other 3 property tests. I had to extend the ExUnit timeout period to make this work, but it was pretty cool to see.

@film42 anything else you think should be changed before merging this? I was thinking I might add a few property tests that exercise an actual `gnat` connection next by subscribing/publishing/etc. And then go back to #69 and test that optimization with the property tests in place for it.

PR #73, opened 2018-05-15
Single request sub

mmmries:
This is a rebase of #69

This code is now being exercised the property tests introduced in.

@film42 was the ideas of `sub + :as_request` to enable custom request/reply flows like the one that MX uses? If so, I might take a stab at an alternate method to enabling those without using the `sub` function. Just want to make sure I understand before suggesting any changes.

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -132,14 +132,14 @@ defmodule Gnat do
   @spec request(GenServer.server, String.t, binary(), keyword()) :: {:ok, message} | {:error, :timeout}
   def request(pid, topic, body, opts \\ []) do
     receive_timeout = Keyword.get(opts, :receive_timeout, 60_000)
-    {:ok, inbox} = new_inbox(pid)
-    {:ok, subscription} = GenServer.call(pid, {:request, %{recipient: self(), inbox: inbox, body: body, topic: topic}})
-    receive do
-      {:msg, %{topic: ^inbox}=msg} -> {:ok, msg}
+    {:ok, subscription} = GenServer.call(pid, {:request, %{recipient: self(), body: body, topic: topic}})
+    response = receive do
+      {:msg, %{topic: ^subscription}=msg} -> {:ok, msg}
       after receive_timeout ->
-        :ok = unsub(pid, subscription)
         {:error, :timeout}
     end
+    :ok = unsub(pid, subscription)
+    response
   end
 
   @doc """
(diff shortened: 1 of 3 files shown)


lib/gnat.ex
```diff
@@ -273,6 +353,15 @@ defmodule Gnat do
     %{state | receivers: receivers}
   end
 
+  defp process_message({:msg, topic, @request_sid, reply_to, body}, state) do
```
film42: How does `@request_sid` work as a method argument? Is it pattern matching on the variable in this module? Is that what we want?

lib/gnat.ex
```diff
@@ -273,6 +353,15 @@ defmodule Gnat do
...
+    if Map.has_key?(state.request_receivers, topic) do
+      send state.request_receivers[topic], {:msg, %{topic: topic, body: body, reply_to: reply_to, gnat: self()}}
+      state
+    else
+      Logger.error "#{__MODULE__} got a response for a request, but that is no longer registered"
```
film42: What do you think about the case where someone publishes a response to a request after we've timed out? Normally we wouldn't receive the message because we've unsub'd, but now that we're using a shared wildcard subscription for all requests, this code path isn't really an error anymore.

Looking at the go-nats client, they drop any message that doesn't match:
1. https://github.com/nats-io/go-nats/blob/d4841f2d6a3ab7745d940768bfbc2fe9d60c65ed/nats.go#L2140-L2164
2. Example showing similar code ignores unknown response messages without raising an error: https://play.golang.org/p/3pVFp3b8O3G

mmmries: :+1: that sounds like a good behavior to me. No reason to log an additional error if we do eventually receive a response, we already would have returned an error indication to the user when the timeout occured

lib/gnat.ex
mmmries: @film42 yes, this function clause will only match if you call it with a tuple that has `:msg` in the first position and `0` (`@request_sid`) in the third position. I think this makes sense because we want to segment any messages we have received for `sid=0` and handle them separately than other messages

film42: For some reason I was thinking `@request_sid` was a value that changed, but it's supposed to be a constant. That makes much more sense :+1:  Thanks for helping me learn :)

mmmries:
@film42 I'm coming back around to this PR (much later) and wondering if you have any opinion on me moving this forward, but without the `as_request` option?

I'm thinking that it's unlikely that MX's use-case will be pressing anytime in the near future since they've gone this long without using it. And if I can leave it out as-of this PR, then I don't run the risk of needing to deprecate/remove it in the future.

PR #74, opened 2018-10-03
require elixir 1.5+, test on last 3 major release of elixir and OTP

mmmries:
Drop support for Elixir `1.4.X` and OTP `18.X`. We'll test against the latest Elixir 1.5, 1.6 and 1.7 releases and the latest minor versions of OTP 19, 20 and 21.


film42:
Looks good!

PR #76, opened 2018-10-11
Fix bug

mmmries:
😭 Found this bug after using `gnat` in a side project. I had a typo in my pattern match which meant that we never matched on the clause where we handle the connection being down and needing to re-connect / re-subscribe.

@jjcarstens @newellista I think you might want to upgrade to `0.5.2` to pickup this fix for `brokaw`


jjcarstens:
😱 spellin iz hard

PR #78, opened 2018-10-30
Expose SID in messages

entone:
I have built a wrapper around the Gnats client to enable instrumentation of inbound and outbound messages. In order to properly correlate the the messages with the topics, and respond correctly I need access to the SID for each subscription.

It's a small change, but I understand it exposes some of the inner workings of Nats. I'd be interested to hear if there are other ways to instrument the Gnat client.

The instrumented client is available here. https://github.com/CityBaseInc/instrumented_gnat


mmmries:
@entone I'm glad to hear you've been using `gnat` and (presumably) it's been useful to you. I'm not at all opposed to sending the `sid` in messages. I didn't include it because I didn't have a use-case for it, but I can't think of a case that we would ever deliver a message that didn't originate from an `sid`.

Let me review this one more time this evening with a little more time and calm. I'll try to read over your example of an instrumented client to see if it generates any ideas. Have you looked at [telemetry](https://hex.pm/packages/telemetry) by any chance? That's the only other thought that I've had about how we could instrument this library directly and then let your code decide what to do with those events?

entone:
I haven't seen Telemetry, seems like it could be quite useful.

As it is, I end wrapping a lot of the calls to Gnat, would definitely prefer to NOT do that :)

mmmries:
I put up an alternative way of getting instrumentation out of `gnat` in #79, but I'm still happy to merge this if it's still useful to you @entone. I don't think it hurts any users of the library or the maintainability of the library.

entone:
I think in general it's not a horrible idea to expose this. My other use case is enforcing a schema on messages eg serializing/deserializing JSON payloads. In general it gives developers the ability to wrap the library for their needs.

entone:
thank you!

entone:
@mmmries think this can get a release as well?

mmmries:
:+1: just released it as `0.5.4` https://hex.pm/packages/gnat/0.5.4

Thanks again for making `gnat` better

PR #79, opened 2018-11-01
add telemetry to the project

mmmries:
@entone this is an alternative to you wrapping the `gnat` library. It should allow you to run something like:

* `Telemetry.attach("record_pubs", [:gnat, :pub], Publish, :instrument, nil)`
* `Telemetry.attach("record_subs", [:gnat, :sub], Subscription, :instrument, nil)`
* `Telemetry.attach("record_requests", [:gnat, :request], Request, :instrument, nil)`
* `Telemetry.attach("record_messages", [:gnat, :message_received], Message, :instrument, nil)`
* `Telemetry.attach("record_unsubs", [:gnat, :unsub], Unsubscribe, :instrument, nil)`

Would that give you all the instrumentation you need? Each of the events have metadata about the `topic` except for the `[:gnat, :unsub]` event because we only have the `sid`.  I also ran the benchmarks script before and after these changes and there was no significant differences.

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -72,7 +72,12 @@ defmodule Gnat do
   ```
   """
   @spec sub(GenServer.server, pid(), String.t, keyword()) :: {:ok, non_neg_integer()} | {:ok, String.t} | {:error, String.t}
-  def sub(pid, subscriber, topic, opts \\ []), do: GenServer.call(pid, {:sub, subscriber, topic, opts})
+  def sub(pid, subscriber, topic, opts \\ []) do
+    start = :erlang.monotonic_time()
+    result = GenServer.call(pid, {:sub, subscriber, topic, opts})
+    Telemetry.execute([:gnat, :sub], :erlang.monotonic_time() - start, %{topic: topic})
+    result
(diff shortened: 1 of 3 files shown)


lib/gnat.ex
```diff
@@ -318,6 +335,7 @@ defmodule Gnat do
   end
   defp process_message({:msg, topic, sid, reply_to, body}, state) do
     unless is_nil(state.receivers[sid]) do
+      Telemetry.execute([:gnat, :message_received], 1, %{topic: topic})
```
jjcarstens: Whats the `1` here? Should there be a latency calculation as well?

jjcarstens: 💯 Fantastic 🎉

entone:
This looks fantastic. I will definitely be migrating my code to use this.

mmmries:
Just released this change as version [0.5.3](https://hex.pm/packages/gnat/0.5.3). Thanks for the great suggestion @entone 🎈 🎉 🎆

entone:
no problem, thanks for tackling this so quickly.

PR #82, opened 2018-12-30
Switch to using NimbleParsec for parsing protocol messages

mmmries:
Parsing isn't a humongous bottleneck right now, but it definitely had some room for improvement. After reading recently about [nimble_parsec](https://hex.pm/packages/nimble_parsec) I thought it might be a good option for efficiently parsing protocol messages.

I wanted to record some better benchmarks before making the upgrade so I went ahead and pulled in the client/server scripts from [recent benchmarking attempts](https://gist.github.com/mmmries/08fe44fdd47a6f8838936f41170f270a) and split out the benchmarks into separate files where we test 16 byte, 128 byte and 1024 byte message sizes.

Here are some of the metrics I found from before/after the change to nimble parsec:

| Metric | Parser | Parsec |
| ------ | ------ | ------- |
| Parse 16 byte ips | 89.35k | 365.37k |
| Parse 128 byte ips | 88.70k | 345.51k |
| Parse 1024 byte ips | 89.11k | 327.81k |
| req/sec 16 byte | 11.1k | 13.1k |
| median req latency | 43.8ms | 36.6ms |

> the throughput metrics were based on running the `server.exs` and `client.exs` on the same 2-core machine with 4 connections and 512 parallel requests


mmmries:
One interesting thing I noticed while working on this was that [the nats protocol](https://nats.io/documentation/internals/nats-protocol/#protocol-conventions) documentation says that subject names should be

> non-empty alphanumeric strings with no embedded whitespace, and optionally token-delimited using the dot character

We currently generate an inbox subjects like `_INBOX.nuid().nuid()`. The `_` is not alphanumeric and the base64 encoded data might include `+`, `/` or `=` characters as well. It looks like the `gnatsd` server handles all of these just fine, so I relaxed the constraint on the new parsing to include pretty much all printable ascii characters other than whitespace and control codes.

Should we avoid non alphanumeric characters in our subject names?

lib/gnat/parsec.ex
```diff
@@ -0,0 +1,104 @@
...
+  import NimbleParsec
+
+  subject = ascii_string([?!..?~], min: 1)
+  length = integer(min: 1)
+  sid = integer(min: 1)
+  whitespace = ascii_char([32, ?\t])
+               |> times(min: 1)
+  op_msg = ascii_char([?m,?M])
+           |> ascii_char([?s,?S])
+           |> ascii_char([?g,?G])
```
josevalim: Curiosity: are those either "MSG" or "msg" or can they be literally "MsG"?

mmmries: That [nats documentation](https://nats.io/documentation/internals/nats-protocol/#protocol-messages) says:

> NATS protocol operation names are case insensitive, thus `SUB foo 1\r\n` and `sub foo 1\r\n` are equivalent

So I went with the literal "case-insensitive" definition, but their example only shows an upcase or downcase 🤷‍♂️

PR #86, opened 2019-02-28
updates to telemetry 0.4

zdcthomas:
Updates telemetry to newest version and removes warnings about deprecated execute/3 with an int 2ng arg. Now uses maps instead

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -161,8 +161,7 @@ defmodule Gnat do
   def unsub(pid, sid, opts \\ []) do
     start = :erlang.monotonic_time()
     result = GenServer.call(pid, {:unsub, sid, opts})
-    latency = :erlang.monotonic_time() - start
-    :telemetry.execute([:gnat, :unsub], %{latency: latency})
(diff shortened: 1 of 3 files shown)


mmmries:
@zdcthomas thanks for the PR ❤️ 💛 💙 💚 

I'm going to take a little time to look at the telemetry updates and think about them. For instance on the `received_message` event, sending `latency: 1` seems a bit odd. Maybe for things like counting events we should instead do `count: 1`? I'll look around to see how other libraries are handling similar use-cases. I also will want to update the `README` with examples of how to listen for telemetry events.

mmmries:
Thanks again for this PR. I'm going to change the one `:telemetry.execute` call to use `count: 1` for its measurements map, and I'll add a few notes to the README about, but otherwise I'll keep all these changes as you wrote them.

mmmries:
Got this merged and then added the couple of tweaks I mentioned above. The changes have been released as `0.6.1` on hex. Thanks again for opening the PR. I'm thrilled that `gnat` has been useful to you. ❤️ 🎉 ⚡️ 💙 🐎

PR #87, opened 2019-03-04
report count measurement, and update README

mmmries:
Builds off of #86

Mostly just some updates to the README, I also changed the measurements for the `message_received` event to be a `count` instead of a `latency`. I'm trying to avoid the flakey auto-unub tests by giving 10ms for the server to finish registering our subscription before we start publishing messages.


mmmries:
Flakey test is not fixed, but everything else looks good

PR #89, opened 2019-04-15
first attempt to setup CircleCI

mmmries:
TravisCI requires me to be an owner of the organization in order to setup builds, and I've been meaning to learn CircleCI for a while now. So I'm changing this project to use CircleCI for its builds. This PR just starts the process of getting those builds running.


mmmries:
@newellista This looks good to go. I got the caching lined up so that builds are fast and we are testing all the same stuff in Circle that we were in Travis. I really liked how Circle splits the checks into 4 different named checks 👌

.circleci/config.yml
```diff
@@ -0,0 +1,93 @@
...
+
+    parallelism: 1
+    docker:
+      - image: elixir:<< parameters.version >>
+
+    working_directory: ~/app
+
+    steps:
+      - checkout  # check out source code to working directory
+      - run: wget https://github.com/nats-io/gnatsd/releases/download/v1.4.1/gnatsd-v1.4.1-linux-amd64.zip
```
film42: Do you want to keep this locked down to a specific version or would you rather run a docker service like `nats:latest`?

mmmries: I looked at using a docker container, but starting 4 of them with different settings was a bit awkward and I like the idea of picking a specific version to test against. I think once 2.0 is released we'll probably test against 2 versions at a time.

PR #90, opened 2019-05-10
Streaming - Experimental Support

mmmries:
This is a ~~first~~ ~~second~~ third pass at streaming support. It follows a pattern of using `get_statem` process that finds a gnat connection, tries to register itself and then monitors the connection so that it can reset the state machine if the connection dies.

For example, here is a state diagram for a subscription

Screen Shot 2019-05-10 at 8 22 29 AM

In future PRs I'm imagining a higher-level supervisor that will automatically setup the connection, streaming client and subscribers by adding a single supervisor to a user's app.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,10 +18,12 @@ jobs:
 
     steps:
       - checkout  # check out source code to working directory
-      - run: wget https://github.com/nats-io/gnatsd/releases/download/v1.4.1/gnatsd-v1.4.1-linux-amd64.zip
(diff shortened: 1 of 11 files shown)


mmmries:
I was feeling stuck with how to release this additional support when I am still very un-sure of whether this is the right API I was users to use, so I've decided to close this PR and start a separate `gnat_streaming` package that will handle all things streaming. That way I can release that as a `0.0.1` version and make it very clear that it is a alpha/beta version of the code

PR #92, opened 2019-09-25
Authentication via token

thousandsofthem:
* Added ability to do auth via token instead of user/pass
* Added README section with auth examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,16 @@ receive do
 end
 ```
 
+## Authentication
+
+```elixir
+# with user and password
+{:ok, gnat} = Gnat.start_link(%{host: '127.0.0.1', port: 4222, username: "joe", password: "123", auth_required: true})
+
+# with token
+{:ok, gnat} = Gnat.start_link(%{host: '127.0.0.1', port: 4222, token: "secret", auth_required: true})
+```
+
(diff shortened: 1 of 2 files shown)


mmmries:
❤️ 💙 💚 💛 ❤️ 💙 💚 💛 ❤️ 💙 💚 💛 ❤️ 💙 💚 💛 
@thousandsofthem thank you so much for contributing to the project
 ❤️ 💙 💚 💛 ❤️ 💙 💚 💛 ❤️ 💙 💚 💛 ❤️ 💙 💚 💛 

I'm wondering if we should add an end-to-end test that just verifies that we can login via token. Something along the lines of the `gnat_test.exs` file's "connect to a server with authentication" test? I haven't setup token authentication before so I'm not sure how hard that is. Have you tested this against a live server?

thousandsofthem:
Yes, this code connects to live server via token auth just fine

mmmries:
Sounds good, I'll try adding in a test really quickly within the master branch and then get this released on hex.pm. Thanks again!

mmmries:
This has been released as `0.7.0` on hex.pm. Thanks again for your work on this

thousandsofthem:
👍

PR #97, opened 2020-01-23
Async Publishing Queue

mmmries:
This is a start at the idea of how to support an async publish queue. Following the patterns from `ConsumerSupervisor` and `ConnectionSupervisor` this idea is started as a separate supervised process. This process manages a `:queue` of messages (we use a `queue` to get O(1) access to the tail of the queue). When you publish a message to the async queue it will start trying to publish that message to the connection with the given  connection name (this pairs nicely with `ConnectionSupervisor`).

### Example Usage

In an application supervisor you would do something like:

```
gnat_supervisor_settings = %{
  name: :gnat, # (required) the registered named you want to give the Gnat connection
  backoff_period: 4_000, # number of milliseconds to wait between consecutive reconnect attempts (default: 2_000)
  connection_settings: [
    %{host: '10.0.0.100', port: 4222},
    %{host: '10.0.0.101', port: 4222},
  ]
}

import Supervisor.Spec

children = [
  worker(Gnat.ConnectionSupervisor, [gnat_supervisor_settings]),
  worker(Gnat.AsyncPub, [%{connection_name: :gnat}, name: :gnat_queue])
]

Supervisor.start_link(children, strategy: :one_for_one)
```

Then throughout your code you could call `Gnat.AsyncPub.pub(:gnat_queue, topic, message, opts)`. The message will be enqueued and delivered when a connection becomes available.

@darrenclark @film42 @jjcarstens any thoughts on this API?

### TODO

- [ ] enforce `max_messages` and `max_memory` settings
  * return `{:error, :queue_full}` or `{:error, :memory_overflow}` if these settings are violated
- [ ] add module docs
- [ ] add function docs
- [x] benchmark overhead of using the async pub vs the synchronous pub
- [x] pause between failed publishes (see issue below)
- [x] implement `stop` so that it waits for the queue to drain
- [x] when supervisors shut down the `AsyncPub`, try to drain the queue first before shutting down
- [ ] timeout in-flight messages that haven't received an ack
- [ ] handle `nack`s from the Gnat connection? (not sure if we'll need this)

### Existing Issues

Currently when we `catch` an error trying to send the publish to the `Gnat` connection, we re-enqueue the message at the beginning of the queue. This makes it so that at least one failing message doesn't block the rest of the queue, but we don't have any sort of backoff or maximum retry limit, so we end up making one scheduler very busy until the message makes it through.

Screen Shot 2020-01-22 at 11 45 01 PM

We should probably build in some sort of retry mechanism that has a maximum number of attempts and  maybe does some sort of backoff to make sure we aren't just spamming the connection or the scheduler.

### DIFF
--- a/lib/gnat/async_pub.ex
+++ b/lib/gnat/async_pub.ex
@@ -3,6 +3,7 @@ defmodule Gnat.AsyncPub do
   require Logger
 
   @default_queue_settings %{max_messages: 10_000, max_memory: 100 * 1024 * 1024}
+  @max_in_flight 10
   @pause 25
 
   def start_link(queue_settings \\ %{}, opts \\ []) do
@@ -22,6 +23,7 @@ defmodule Gnat.AsyncPub do
       @default_queue_settings
       |> Map.merge(queue_settings)
       |> Map.put(:queue, :queue.new())
+      |> Map.put(:in_flight, %{})
     {:ok, queue_settings, 0}
   end
 
@@ -40,26 +42,43 @@ defmodule Gnat.AsyncPub do
     if :queue.is_empty(state.queue) do
       {:noreply, state, @pause}
     else
-      new_queue = try_to_send_message(state)
-      state = Map.put(state, :queue, new_queue)
-      {:noreply, state, 0}
+      {state, timeout} = try_to_send_message(state)
+      {:noreply, state, timeout}
     end
   end
+  def handle_info({ref, :ok}, state) do
+    in_flight = Map.delete(state.in_flight, ref)
+    {:noreply, %{state | in_flight: in_flight}, 0}
+  end
+  def handle_info(other, state) do
+    Logger.error("#{__MODULE__} Received Unexpected Message: #{inspect(other)}")
+    {:noreply, state}
+  end
 
-  defp try_to_send_message(%{queue: queue, connection_name: name}) do
+  defp try_to_send_message(%{queue: queue, connection_name: name, in_flight: in_flight}=state) do
     # :queue.is_empty is checked above
-    {{:value, {:pub, topic, message, opts} = todo}, queue} = :queue.out(queue)
-    case attempt_pub(name, topic, message, opts) do
-      :ok -> queue
-      :error -> :queue.in(todo, queue)
+    {{:value, pub}, new_queue} = :queue.out(queue)
+    case Enum.count(in_flight) do
+      n when n < @max_in_flight ->
+        case attempt_pub(name, pub) do
+          {:ok, ref} ->
+            in_flight = Map.put(in_flight, ref, {:erlang.monotonic_time(), pub})
+            {%{state | in_flight: in_flight, queue: new_queue}, 0}
+          :connection_down ->
+            {state, @pause}
+        end
+      _ -> {state, @pause}
     end
   end
 
-  defp attempt_pub(connection_name, topic, message, opts) do
+  defp attempt_pub(connection_name, pub) do
+    ref = :erlang.make_ref()
+    process_message = {:"$gen_call", {self(), ref}, pub}
     try do
-      Gnat.pub(connection_name, topic, message, opts)
-    catch
-      :exit, _msg -> :error
+      :ok = Process.send(connection_name, process_message, [])
+      {:ok, ref}
+    rescue
+      ArgumentError -> :connection_down
     end
   end
 end
--- a/lib/gnat/async_pub.ex
+++ b/lib/gnat/async_pub.ex
@@ -5,42 +5,48 @@ defmodule Gnat.AsyncPub do
   @default_queue_settings %{max_messages: 10_000, max_memory: 100 * 1024 * 1024}
   @max_in_flight 10
   @pause 25
+  @shutting_down_status :shutting_down
 
   def start_link(queue_settings \\ %{}, opts \\ []) do
     GenServer.start_link(__MODULE__, queue_settings, opts)
   end
 
-  @spec pub(GenServer.server, String.t, binary(), keyword()) :: :ok
+  @type pub_error :: :shutting_down | :queue_full | :memory_full
+  @spec pub(GenServer.server, String.t, binary(), keyword()) :: :ok | {:error, pub_error()}
   def pub(pid, topic, message, opts \\ []) do
     GenServer.call(pid, {:pub, topic, message, opts})
   end
 
-  def stop(pid), do: GenServer.call(pid, :stop)
-
   @impl GenServer
   def init(%{connection_name: name}=queue_settings) when is_atom(name) do
+    Process.flag(:trap_exit, true)
+    Process.flag(:trap_exit, true)
     queue_settings =
       @default_queue_settings
       |> Map.merge(queue_settings)
       |> Map.put(:queue, :queue.new())
       |> Map.put(:in_flight, %{})
+      |> Map.put(:status, :running)
     {:ok, queue_settings, 0}
   end
 
   @impl GenServer
+  def handle_call({:pub, _top, _msg, _opts}, _from, %{status: @shutting_down_status}=state) do
+    {:reply, {:error, :shutting_down}, state, 0}
+  end
   def handle_call({:pub, _topic, _message, _opts}=todo, _from, state) do
     state = update_in(state, [:queue], fn(q) -> :queue.in(todo, q) end)
     {:reply, :ok, state, 0}
-  end
-  def handle_call(:stop, _from, state) do
-    # wait for queue to empty?
-    {:stop, :normal, :ok, state}
   end
 
   @impl GenServer
   def handle_info(:timeout, state) do
     if :queue.is_empty(state.queue) do
-      {:noreply, state, @pause}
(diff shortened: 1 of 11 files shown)


lib/gnat/async_pub.ex
```diff
@@ -0,0 +1,59 @@
...
+    {{:value, {:pub, topic, message, opts} = todo}, queue} = :queue.out(queue)
+    case attempt_pub(name, topic, message, opts) do
+      :ok -> queue
+      :error -> :queue.in(todo, queue)
+    end
+  end
+
+  defp attempt_pub(connection_name, topic, message, opts) do
+    try do
+      Gnat.pub(connection_name, topic, message, opts)
```
film42: Does each call to pub include a call to flush? Just wondering if we can pipeline a batch of messages to Nats sitting in the queue.

mmmries: The `Gnat` connection handles some pipelining ([see these lines](https://github.com/nats-io/nats.ex/blob/dd755d1d6e963c8c8ef402d11373727f84ccf511/lib/gnat.ex#L255-L257)), but it is assuming that multiple different callers might be sending those pub messages.

It's probably worth adding something in this process which attempts to send multiple pub messages to the connection and then waits for the `acks` before deciding to re-queue them or throw them out.

mmmries: I'll try to get a benchmark in place so we have some idea of what the performance difference is

lib/gnat/async_pub.ex
```diff
@@ -0,0 +1,59 @@
...
+      @default_queue_settings
+      |> Map.merge(queue_settings)
+      |> Map.put(:queue, :queue.new())
+    {:ok, queue_settings, 0}
+  end
+
+  @impl GenServer
+  def handle_call({:pub, _topic, _message, _opts}=todo, _from, state) do
+    state = update_in(state, [:queue], fn(q) -> :queue.in(todo, q) end)
+    {:reply, :ok, state, 0}
```
jjcarstens: I think using this `0` timeout for flow will probably work fine but I'd also suggest using `:continue` instead for these immediate actions based on what the GenServer docs say:
> _Because a message may arrive before the timeout is set, even a timeout of 0 milliseconds is not guaranteed to execute. To take another action immediately and unconditionally, use a :continue instruction._

mmmries: I actually wrote it using `:continue` and `handle_continue` initially, but then I realized that when someone does an async pub, the process would immediately become temporarily un-available until the `handle_continue` finishes.

If there are a bunch of `pub`s right at the same time, I think I would prefer to be responsive those processes and fit in the next push of messages to the connection right after them... I guess this means that it's possible for the process to be so busy responding to `pub`s that it never actually gets anything published 🤔 I think that would be pretty hard to do though.

jjcarstens: Seems like a fair trade-off.

We might be able to resolve that by using `cast` instead (unless you want `call` to guarantee the message got updated to the queue). Then could use the `:continue` flow.

But...I think its 6's at this point and can be addressed should that crazy usage case even come up. Totally good with implementation and just spitballing other thoughts

mmmries: :+1: the `cast` idea interesting for sure. I'll think about that one

mmmries:
I've added an async publishing  benchmark that measures the difference between synchronous and asynchronous publishing. This benchmark measures the median and 99% latency of receiving the published messages and also the effective throughput under varying numbers of publishers. 

```
riespro:gnat$ mix run bench/async_pub.exs 
## 10 publishers
        Sync
                Completed 100000 messages in 618.577ms
                161661 msg/sec, median: 271ms, 99%: 453ms
        Async
                Completed 100000 messages in 1379.884ms
                72470 msg/sec, median: 667ms, 99%: 1025ms
## 100 publishers
        Sync
                Completed 100000 messages in 672.453ms
                148709 msg/sec, median: 316ms, 99%: 385ms
        Async
                Completed 100000 messages in 1286.958ms
                77703 msg/sec, median: 652ms, 99%: 1100ms
## 1000 publishers
        Sync
                Completed 100000 messages in 1894.696ms
                52779 msg/sec, median: 1656ms, 99%: 1853ms
        Async
                Completed 100000 messages in 1290.23ms
                77506 msg/sec, median: 656ms, 99%: 1094ms
```

mmmries:
I've added an async publishing benchmark that measures the difference between synchronous and asynchronous publishing. This benchmark measures the median and 99% latency of receiving the published messages and also the effective throughput under varying numbers of publishers.

```
## 1 publishers
        Sync
                Completed 500000 messages in 4694.773ms
                106501 msg/sec, median: 50µs, 99%: 160µs, min: 23µs
        Async
                Completed 500000 messages in 8383.9ms
                59638 msg/sec, median: 68µs, 99%: 134µs, min: 29µs
## 10 publishers
        Sync
                Completed 500000 messages in 3296.634ms
                151670 msg/sec, median: 656557µs, 99%: 1310246µs, min: 111µs
        Async
                Completed 500000 messages in 6765.027ms
                73910 msg/sec, median: 3212844µs, 99%: 5259710µs, min: 267µs
## 100 publishers
        Sync
                Completed 500000 messages in 3706.586ms
                134895 msg/sec, median: 1353609µs, 99%: 2138161µs, min: 398µs
        Async
                Completed 500000 messages in 6517.001ms
                76722 msg/sec, median: 3197926µs, 99%: 5575893µs, min: 68243µs
## 1000 publishers
        Sync
                Completed 500000 messages in 4183.393ms
                119520 msg/sec, median: 1142722µs, 99%: 1717775µs, min: 394µs
        Async
                Completed 500000 messages in 6853.453ms
                72956 msg/sec, median: 3277064µs, 99%: 5808468µs, min: 1026407µs
```

mmmries:
Pushed up a change where we manufacture the `GenServer.call` messages and receive the responses concurrently. This allows us to pipeline our pubs so we spend less time blocking the queue process. This gives us 38% better throughput than synchronous calls for a single publishing process and 12% worth throughput in the optimal case of having 10 publishers.

```
## 1 publishers
        Sync
                Completed 500000 messages in 4841.827ms
                103267 msg/sec, median: 49µs, 99%: 160µs, min: 22µs
        Async
                Completed 500000 messages in 3492.171ms
                143177 msg/sec, median: 136µs, 99%: 517µs, min: 33µs
## 10 publishers
        Sync
                Completed 500000 messages in 3172.216ms
                157619 msg/sec, median: 331957µs, 99%: 988021µs, min: 66µs
        Async
                Completed 500000 messages in 3547.411ms
                140948 msg/sec, median: 1510908µs, 99%: 2058566µs, min: 269µs
## 100 publishers
        Sync
                Completed 500000 messages in 3702.223ms
                135054 msg/sec, median: 1291986µs, 99%: 2233458µs, min: 482µs
        Async
                Completed 500000 messages in 3540.837ms
                141210 msg/sec, median: 1759578µs, 99%: 2598027µs, min: 203946µs
## 1000 publishers
        Sync
                Completed 500000 messages in 4125.113ms
                121209 msg/sec, median: 1043357µs, 99%: 1714589µs, min: 416µs
        Async
                Completed 500000 messages in 3967.74ms
                126016 msg/sec, median: 2109741µs, 99%: 2915924µs, min: 1071588µs
```

mmmries:
Hey @ColinSullivan1 I heard though the grapevine that you are going to be the core maintainer over-seeing the `nats.ex` project. Just wanted to let you know that this PR is under active development and see if you have any input on the structure or API?

mmmries:
Got a version of this working that has graceful shutdown (a slow version that always waits 5 seconds). The benchmarks are looking mostly the same with these changes in place.

```
## 1 publishers
        Sync
                Completed 500000 messages in 4734.297ms
                105612 msg/sec, median: 50µs, 99%: 164µs, min: 23µs
        Async
                Completed 500000 messages in 3547.68ms
                140937 msg/sec, median: 143µs, 99%: 499µs, min: 33µs

## 10 publishers
        Sync
                Completed 500000 messages in 2127.926ms
                234971 msg/sec, median: 549µs, 99%: 7159µs, min: 50µs
        Async
                Completed 500000 messages in 2792.566ms
                179047 msg/sec, median: 1303373µs, 99%: 1412058µs, min: 239µs

## 100 publishers
        Sync
                Completed 500000 messages in 3793.089ms
                131819 msg/sec, median: 1345784µs, 99%: 2277790µs, min: 988µs
        Async
                Completed 500000 messages in 2627.372ms
                190304 msg/sec, median: 1352415µs, 99%: 1646713µs, min: 81759µs

## 1000 publishers
        Sync
                Completed 500000 messages in 3906.09ms
                128005 msg/sec, median: 919704µs, 99%: 1671042µs, min: 994µs
        Async
                Completed 500000 messages in 2513.735ms
                198907 msg/sec, median: 1283260µs, 99%: 1580543µs, min: 991482µs
```

mmmries:
This PR has gotten pretty stale and I still haven't had a use-case for this in a real application so I'm going to close the PR. If anyone finds this interesting feel free to comment here and start a discussion on why this should be re-considered

PR #105, opened 2020-08-27
change log from info to debug when connecting to NATS

### LINKED ISSUE #104: Connection establishing prinout
Hi,

Currently the connection prinout here https://github.com/nats-io/nats.ex/blob/f2accd13a65c751543cb44df367ac43fd93071ae/lib/gnat/connection_supervisor.ex#L60 uses `Logger.info`, this means that if you are deploying this with authentication enabled you will print the secrets into std.out unless you are silencing all info messages. 

I have the following proposals:
- change it to Logger.debug

OR

- strip any secret data from this printout so that the URL, PORT, AUTH_REQUIRED and any other non-secret info gets printed

Thanks for any other suggestions or help :) We're already using this in production for one of our apps.

fallion:
Currently the ConnectionSupervisor will print out all inputs as part of Logger info. This can lead to unintended leaking of credentials into std.out

Closes #104

### DIFF
--- a/lib/gnat/connection_supervisor.ex
+++ b/lib/gnat/connection_supervisor.ex
@@ -57,7 +57,7 @@ defmodule Gnat.ConnectionSupervisor do
   @impl GenServer
   def handle_info(:attempt_connection, state) do
     connection_config = random_connection_config(state)
-    Logger.info "connecting to #{inspect connection_config}"
+    Logger.debug "connecting to #{inspect connection_config}"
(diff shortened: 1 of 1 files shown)


mmmries:
@fallion this change was released as version `1.1.3` on hexpm https://hex.pm/packages/gnat/1.1.3

fallion:
@mmmries thanks! I'll look into other issues I could help out with.

PR #111, opened 2021-01-06
Create dependencies.md [ci skip]

gcolliso:


mmmries:
@gcolliso is this a file that I should start maintaining as part of the project? I think it would be helpful to break out the runtime dependencies vs the development dependencies so that people using the project know which of those licenses affect them?

gcolliso:
@mmmries If you would maintain this file, that would be great, but is not necessary. We are putting dependency files in each nats-io org supported clients with a link to them from the nats-io/general repo. If you would like to organize them differently,  you are welcome to. Thanks.

mmmries:
@gcolliso thanks for the clarification. I went ahead and broke the file into two sections to make it clear which dependencies would be pulled into project when people use this library.

Issue #115, opened 2021-06-09
add Headers Support


byu:
Nats 2.2+ has added new protocol for publishing/subscribing messages w/ http-like headers.
it would be nice for this client lib to support this new protocol.

mmmries:
💯 I would love to add support for headers here. I'm planning to add it before too long, but if anyone else wants to take a crack at it first I would take a look at the `lib/gnat/parsec.ex` file which we use to parse the raw TCP bytes into message structs.

mmmries:
I started to look into this and @wallyqs pointed me to https://github.com/nats-io/nats-server/blob/be9fb5f6f2455c40691260986bdb37b1e5796046/doc/adr/0004-nats-headers.md which documents how to support headers.

I'll try to take a crack at this soon, but if anyone wants to start on it before I do, I can help point you in the right direction

autodidaddict:
Just came here to second the need/desire for headers as I've got Go and Rust clients attempting to send my Elixir code messages with headers.

mmmries:
Hey @autodidaddict any chance you can try out the `1.3.0-rc0` release to see if that matches your needs? I'm leaning towards releasing those changes, but was just wanting to wait for some feedback

autodidaddict:
Looking at the [docs](https://hexdocs.pm/gnat/1.3.0-rc0/Gnat.html#t:message/0) for the `message()` type for 1.3.0-rc0 doesn't show any header support. Is there somewhere else I should be looking for it?

autodidaddict:
@mmmries also looking at the source code and not finding out how to set/read the headers on message publications through `request` or through `pub`. This is likely due to me not understanding what I'm looking for/at, but if you could point me at some sample code that uses headers in both directions, that would be amazing.

mmmries:
@autodidaddict I haven't written up changes to the docs yet, but here is what changes for `1.3.0-rc0`

* headers are represented as a list of tuples (this allows repeated keys). `[{"header name", "header value"}, ...]`
* when receiving a message that was sent with headers, you'll get a map that includes a `:headers` key a list of headers, this key will only be present for messages published with headers, it is omitted for normal messages
* When calling `pub` you can pass an optional `:headers` key in the`opts` 4th argument
* When calling `request` you can pass an optional `:headers` key in the `opts` 4th argument

autodidaddict:
Awesome that sounds like exactly what we've want to do with headers

mmmries:
[...]

PR #116, opened 2021-06-13
Support headers

mmmries:
Addresses #115 

This PR will add support for sending and receiving messages with headers. This was added [as part of the nats 2.2 relase](https://docs.nats.io/whats_new_22#message-headers) and allows for providing HTTP-style headers with messages. In order to handle parsing these headers correctly, I'm pulling in the [cowlib](https://hex.pm/packages/cowlib) library as a dependency. This library is heavily used within the Elixir community by the `cowboy` which is part of the standard stack for Phoenix projects. It has 28M downloads which is a lot for the Elixir community 😅 

Parsing messages with headers is of course a little slower than messages without headers, and I added a benchmark which indicates that it's around 3x slower to parse messages with headers compared to those without them. I think this is fine given that the nats team explicitly talks about headers having overhead and the fact that they are totally optional.

```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
Number of Available Cores: 12
Available memory: 16 GB
Elixir 1.12.0
Erlang 24.0

Benchmark suite executing with the following configuration:
warmup: 2 s
time: 10 s
memory time: 0 ns
parallel: 1
inputs: 1024 byte, 128 byte, 16 byte, 7 byte with headers
Estimated total run time: 48 s

Benchmarking parsec with input 1024 byte...
Benchmarking parsec with input 128 byte...
Benchmarking parsec with input 16 byte...
Benchmarking parsec with input 7 byte with headers...

##### With input 1024 byte #####
Name             ips        average  deviation         median         99th %
parsec      753.96 K        1.33 μs  ±2453.84%        0.99 μs        1.99 μs

##### With input 128 byte #####
Name             ips        average  deviation         median         99th %
parsec      740.40 K        1.35 μs  ±3117.98%        0.99 μs        1.99 μs

##### With input 16 byte #####
Name             ips        average  deviation         median         99th %
parsec      723.76 K        1.38 μs  ±3288.75%        0.99 μs        1.99 μs

##### With input 7 byte with headers #####
Name             ips        average  deviation         median         99th %
parsec      257.39 K        3.89 μs   ±973.12%        2.99 μs        7.99 μs
```


mmmries:
I'm opening this in draft mode because I still need to

* decide on an interface for sending messages with headers (this will probably affect both `pub` and `request` functions)
* add code for generating hpub messages
* add code to process the `:hmesg` parsed messages (send this as a map with a new `:headers` key)
* add an end-to-end test for sending/receiving messages with headers
* update the documentation to call out headers

mmmries:
@film42 @newellista I know neither of you have been involved in this project recently, but I wanted to ask your opinion on using the `:cowlib` dependency for parsing and encoding headers. I'm reluctant to add a new dependency, but it seems that HTTP-style headers have some oddities to them so it I'm leaning toward using a library with and established presence in the community rather than re-implementing that myself.

The dependency has no transitive dependencies, and it seems to be a pretty small and well curated set of erlang code.

film42:
Because it's a cowboy dependent lib, I think `cowlib` is perfectly fine to use.

lib/gnat/command.ex
```diff
@@ -1,11 +1,26 @@
...
   @unsub "UNSUB"
 
   def build(:pub, topic, payload, []), do: [@pub, " ", topic, " #{IO.iodata_length(payload)}", @newline, payload, @newline]
   def build(:pub, topic, payload, [reply_to: reply]), do: [@pub, " ", topic, " ", reply, " #{IO.iodata_length(payload)}", @newline, payload, @newline]
+  def build(:pub, topic, payload, [headers: headers]) do
+    # it takes 10 bytes to add the nats header version line
+    # and 2 more for the newline between headers and payload
+    header_len = IO.iodata_length(headers) + 12
+    total_len = IO.iodata_length(payload) + header_len
+    [@hpub, " ", topic, " ", Integer.to_string(header_len), " ", Integer.to_string(total_len), "\r\nNATS/1.0\r\n", headers, @newline, payload, @newline]
```
film42, line 15: Not super clear so just adding a comment. Is the `headers` param already stringified? i.e. `"Header 1: A\r\nHeader 2: B\r\n"`. I'm just wondering if the output needs to be `...header, @newline, @newline...`. I see there are parse tests, but is there a build test that asserts the string form of this output?

mmmries: The `headers` param does get encoded before this (we actually encode it to `iodata` rather than a string to avoid extra allocations), but you can see this in the `def pub` and `def request` methods where we call the `prepare_headers` helper function. I did it this way so that any errors with the headers input (ie passing the wrong kinds of arguments, etc) will happen in the client process, rather than the connection process. This also minimizes the amount of time we spent on encoding in the connection process.

I don't have a test for this yet, but maybe I should add one? I ended up writing the end-to-end test which does a full round trip through a nats server and once that was working I didn't think about adding a unit test for this 🤔

film42:
```
iex(13)> Gnat.Command.build(:pub, "yolo.topic", "intel inside, not", headers: :cow_http.headers([{"apple", "m1"}]), reply_to: "me")
[
  "HPUB",
  " ",
  "yolo.topic",
  " ",
  "me",
  " ",
  "23",
  " ",
  "40",
  "\r\nNATS/1.0\r\n",
  [["apple", ": ", "m1", "\r\n"]],
  "\r\n",
  "intel inside, not",
  "\r\n"
]
```

I just wasn't sure what `:cow_http.headers` responded with (whether the terminating newline was added or not). It is, so 👍 . Just wasn't clear from github code review.

mmmries:
Hey @byu, I released version `1.3.0-rc0` with support for headers. I'll be adding docs for how to use-them before merging and releasing `1.3.0`, but wanted to let you know in case you want to try it out ahead of the release. I'll probably pull this into a production environment at my work to test it out and keep an eye out for any surprises.

mmmries:
My early testing has all been fine on this and I haven't heard any negative feedback so I'm gonna merge this as-is and then I'll add a few commits to the master branch to improve documentation about how to interact with headers and publish this as `1.3.0`

PR #121, opened 2022-03-10
Allow setting custom inbox prefix

mkaput:
This PR adds an ability to customize the hardcoded `_INBOX.` prefix used in some subjects.

In our system, applications get third party JWT tokens allow operating only on specified topics (allow specifier looks usually like this: `app_name.environment.>`). This little option would allow us to use Gnat in such conditions.

I have no idea how to reliably test this parameter, hence no tests are added. If you have any clever ideas, let me know and I can implement these.

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -31,6 +31,7 @@ defmodule Gnat do
     connection_timeout: 3_000,
     ssl_opts: [],
     tls: false,
+    inbox_prefix: "_INBOX.",
   }
 
   @request_sid 0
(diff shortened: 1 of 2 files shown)


film42:
Looks like `inbox_prefix` matches the nats.go option name as well: https://github.com/nats-io/nats.go/blob/8045a5037b0128b90a293579d9d66de6c848c5f2/nats.go#L463-L464

mmmries (commented):
I think there are a couple of tests in `test/gnat_test.exs` that specifically test using the request/reply  pattern. If you can copy one of those tests and pass in this option, that should give us some coverage for being able to pass in the option.

We could assert that the request was sent with a `reply_to` that has your customized inbox prefix.

mkaput:
> Looks like `inbox_prefix` matches the nats.go option name as well: https://github.com/nats-io/nats.go/blob/8045a5037b0128b90a293579d9d66de6c848c5f2/nats.go#L463-L464

Yup, that's where I took the name from 😃

mmmries (approved):
:shipit: This is a great addition! Thanks for opening the PR ❤️ 💛 💙 💚 💜

mkaput:
@mmmries we are going to test this feature from commit sha, though may I kindly ask you to prepare new release with this change? 😃

mmmries:
I've got a commit prepared to bump to `1.5.0` with a note added to the CHANGELOG.md. If you can drop a note about how your testing goes I will release 👍

mmmries:
🎉 Released as gnat `1.5.0` https://hex.pm/packages/gnat/1.5.0

mkaput:
Thank you ♥️

Issue #122, opened 2022-05-11
Idle Heartbeat Header Parsing


brandynbennett:
When working with JetStream Push Consumers there's a type of header message that gets published that has headers like this: `NATS/1.0 100 Idle Heartbeat\r\nNats-Last-Consumer: 2\r\nNats-Last-Stream: 2\r\n`. These messages result in a `{:error, "Could not parse headers"}`. I believe because this line https://github.com/nats-io/nats.ex/blob/main/lib/gnat/parsec.ex#L116 is expecting a line break immediately after the `NATS/1.0` declaration.

mmmries:
I merged #125 and released it as version 1.5.2 on hex.pm

Based on our conversation in that PR, we may still want to have a way of returning the status code from the header line. So I think we can leave this issue open and come back to it when we have a good idea of how to handle that scenario.

brandynbennett:
That sounds great. Thanks!

mmmries:
We haven't found the need to make more modifications here, so I'll go ahead and close this

PR #125, opened 2022-05-16
parse hmsg with status code and description

mmmries:
This fixes the bug from #122

I got the parsec working correctly here, but I think we should also return the status code and/or description back to the recipient. This PR currently throws that data away, but especially in the case of the pull request with no wait, it seems very useful to let the caller know that we got back a 404 vs a 503 or some other status code.

### DIFF
--- a/lib/gnat/parsec.ex
+++ b/lib/gnat/parsec.ex
@@ -113,14 +113,20 @@ defmodule Gnat.Parsec do
     end
   end
 
-  def parse_headers("NATS/1.0\r\n" <> headers) do
-    case :cow_http.parse_headers(headers) do
-      {parsed, ""} -> {:ok, parsed}
-      _other -> {:error, "Could not parse headers"}
+  def parse_headers("NATS/1.0" <> rest) do
+    case String.split(rest, "\r\n", parts: 2) do
+      [_status_line, headers] ->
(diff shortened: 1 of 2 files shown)


brandynbennett:
I agree we should include the status code somehow. I'm less confident about where in the return value to add that

brandynbennett:
[This line from the architecture docs](https://github.com/nats-io/nats-architecture-and-design/blob/682d5cd5f21d18502da70025727128a407655250/adr/ADR-13.md) makes me think the error status should be handled differently. 

> Whenever there is a status message, it is recommended that the error is not returned to the user as a processable message and instead handle is as the error condition

I think that is a separate issue though from https://github.com/nats-io/nats.ex/issues/122 and would require a deeper understanding of possible status codes to parse them and handle them appropriately. This PR at least gets the headers parsing without breaking. It makes sense to me to address the status code handling in separate PR/issue

mmmries:
@brandynbennett thanks for the link to that ADR document. I wasn't thinking of a status code of 100 as an error-case, but I'll go back and read that in it's entirety to see if it makes more sense to me.

I agree though, that we should probably ship this as-is and then come back to the handling of the status code. I think that's what you were saying?

brandynbennett:
Ya I think this PR is good to go, and the status code handling can be a separate PR

PR #128, opened 2022-07-22
Add function to get a new inbox

brandynbennett:
Primarily useful for libraries like Jetstream and Polyn, built on top of `gnat` that also need to setup a temporary inbox to receive messages.

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -363,7 +363,7 @@ defmodule Gnat do
       {:ok, socket, server_info} ->
         parser = Parsec.new
 
-        request_inbox_prefix = Map.fetch!(connection_settings, :inbox_prefix) <> "#{nuid()}."
+        request_inbox_prefix = inbox_name(connection_settings) <> "."
 
         state = %{socket: socket,
                   connection_settings: connection_settings,
@@ -428,7 +428,7 @@ defmodule Gnat do
     {:noreply, state}
   end
   def handle_call({:request, request}, _from, state) do
-    inbox = inbox_name(state)
+    inbox = state.request_inbox_prefix <> nuid()
     new_state = %{state | request_receivers: Map.put(state.request_receivers, inbox, request.recipient)}
     pub =
       case request do
@@ -472,7 +472,7 @@ defmodule Gnat do
     {:reply, state.server_info, state}
   end
   def handle_call(:make_new_inbox, _from, state) do
-    {:reply, inbox_name(state), state}
+    {:reply, inbox_name(state.connection_settings), state}
   end
 
   defp create_request_subscription(%{request_inbox_prefix: request_inbox_prefix}=state) do
@@ -483,7 +483,7 @@ defmodule Gnat do
     add_subscription_to_state(state, @request_sid, self())
   end
 
-  defp inbox_name(%{request_inbox_prefix: prefix}), do: prefix <> nuid()
+  defp inbox_name(%{inbox_prefix: prefix}), do: prefix <> nuid()
 
   defp nuid(), do: :crypto.strong_rand_bytes(12) |> Base.encode64
 
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -348,6 +348,14 @@ defmodule Gnat do
     GenServer.call(name, :server_info)
   end
 
+  @doc """
+  Get a randomly generated subscription inbox based on your inbox_prefix.
+  """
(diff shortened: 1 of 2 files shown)


mmmries:
@brandynbennett this PR looks fine on it's own, but I'm not sure how it will be used. With the normal inbox prefix, the gnat process can decide which replies are being used by which processes (ie who to forward each message to), but it looks like in this case we just generate a subject prefixed by our inbox, but it's not associated with the process that will need to be notified

brandynbennett:
@mmmries That's right, this function does not associate the inbox with any process. My intent here was only to expose the inbox name generation. JetStream and Polyn both have places where they need a temporary inbox and have to duplicate `nuid` creation as well as passing in the custom inbox_prefix. This removes that duplication so that libraries can use the inbox_prefix passed into a `Gnat` connection and not have to configure it again

mmmries:
@brandynbennett my apologies, but I'm still not seeing how the whole use-case comes together. Polyn or Jetstream ask for a temporary inbox, then they ask to receive messages on that inbox? Wouldn't that require you to call `sub` with that new inbox? or how does the Gnat process know to forward those messages to Polyn/Jetstream?

lib/gnat.ex
```diff
@@ -348,6 +348,14 @@ defmodule Gnat do
     GenServer.call(name, :server_info)
   end
 
+  @doc """
+  Get a randomly generated subscription inbox based on your inbox_prefix.
+  """
+  @spec make_new_inbox(t()) :: binary()
+  def make_new_inbox(pid) do
+    GenServer.call(pid, :make_new_inbox)
```
film42, line 356: Because this just returns a string with no other IO involved, it seems like this would be more beneficial to not call through a genserver?

https://github.com/nats-io/nats.ex/blob/a01fe4402f5b0d8f7c3db9e4ea45596693bc6baf/lib/gnat.ex#L475-L477

It really is like 2 lines of code for a random new nats-style inbox string.

mmmries:
Hmmmmmm, have you tried to run end-to-end tests with those other libraries utilizing this technique? I think sharing an inbox prefix with the gnat connection could actually cause some confusion within this library.  I think if jetstream for instance called `sub` with a subject that shares it's inbox prefix with the gnat processe's inbox prefix, we would end up receiving two `MSG`s from the nats cluster. This is because we have two active subscriptions that share a prefix:

* Gnat would be subscribed to `_INBOX.ABC.*`
* Jetstream has added a subscription to `_INBOX.ABC.SomethingElse`

When jetstream publishes a message to `_INBOX.ABC.SomethingElse` we would receive two different messages from the nats cluster, but the first subscription wouldn't know what to do with the message since the "SomethingElse" part doesn't correlate to any of it's request/reply messages that it's tracking.

Overall, I think the pattern of using an inbox prefix is a common pattern in the nats ecosystem and the inbox prefix used by the `Gnat` connection is really meant to support it's synchronous request-reply interface. It's not meant to be a general purpose inbox prefix.

brandynbennett:
Ok, so I didn't realize Gnat was doing a `sub` on `_INBOX.ABC.*` in it's `init`. For a `request_inbox_prefix`. So every `request` call actually gets two `nuid` attached to it. Ya I wasn't wanting to share the `request_inbox_prefix` of `_INBOX.ABC.`. I was only wanting to share the pattern of `:inbox_prefix <> nuid()`. Makes sense why this was so confusing.

mmmries:
@brandynbennett I just came back around to this PR and I have a few questions about how to move it forward:

1. I don't have a performance concern since this will be called very few times by other use-cases like Jetstream
2. I think the intention is to keep the same inbox prefix for request/reply patterns as well as other usages like subscribing to a jetstream stream?
  a. Are there cases where someone would want to use 2 different inbox prefixes?
  b. Is it a lot of friction to add a configurable `inbox_prefix` to jetstream?

This is a pretty small change and I don't have a strong objection, but I want to make sure I understand the use-case. It seems to me like jetstream making it's own temporary inbox is not a lot of duplication and the sharing could be either desirable or undesirable depending on the situation (ie if someone configures gnat with an inbox_prefix because they are wanting to monitor req-reply usage, but now they would be seeing other stuff with the same pattern?)

brandynbennett:
@mmmries I think those are good points. The friction is not very high and it could be the case the a user would want 2 different inbox_prefixes for the different activities. It may be best to close this for now and revisit in the future if the need arises. Thanks for working through this with me.

PR #134, opened 2022-10-06
Support OK message and verbose mode

### LINKED ISSUE #133: Support OK message
From spec https://docs.nats.io/reference/reference-protocols/nats-protocol#protocol-messages

I've had to do [a quick hack](https://github.com/nats-io/nats.ex/commit/0c79e7806a70cd7c3cc7dcdcde700a6e6cca7ed3) to support logs from Fly.io so maybe this is a good thing to add here too?

Would be glad to follow up with a PR

lubien:
Close  #133

### DIFF
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -13,7 +13,7 @@ case :gen_tcp.connect('localhost', 4222, [:binary]) do
 end
 
 # this is used by some property tests, see test/gnat_property_test.exs
-Gnat.start_link(%{}, [name: :test_connection])
+Gnat.start_link(%{verbose: true}, [name: :test_connection])
 
 defmodule RpcEndpoint do
   def init do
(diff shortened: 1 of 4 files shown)


test/test_helper.exs
```diff
@@ -13,7 +13,7 @@ case :gen_tcp.connect('localhost', 4222, [:binary]) do
 end
 
 # this is used by some property tests, see test/gnat_property_test.exs
-Gnat.start_link(%{}, [name: :test_connection])
+Gnat.start_link(%{verbose: true}, [name: :test_connection])
```
mmmries, line 16: Why did we want this connection to run in verbose mode? I would think that because we are using this test for some property tests, we would want it to have minimal overhead?

mmmries (commented):
This PR looks great! I had just the one question about the connection that is used for some shared tests, but other than that I think this is ready to merge

PR #135, opened 2022-10-06
Accept user enforced auth

### LINKED ISSUE #132: Let nats client send auth_required without waiting for the server to ask for it
I've been using a [fork of your library](https://github.com/nats-io/nats.ex/compare/main...lubien:nats.ex:main) on [Fly.io for logs](https://community.fly.io/t/fly-logs-over-nats/1540) with a few changes.

My proposal negotiate auth if the client wrote those on its config without needing for the server to ask for it. I've implemented here an example: https://github.com/nats-io/nats.ex/commit/9ddb11acbca7727e97c6624425a2cb22243cfafe

This feels like a breaking change for some so I'd love input on this!

lubien:
Close #132

### DIFF
--- a/lib/gnat/handshake.ex
+++ b/lib/gnat/handshake.ex
@@ -15,8 +15,10 @@ defmodule Gnat.Handshake do
   end
 
   def negotiate_settings(server_settings, user_settings) do
+    auth_required = server_settings[:auth_required] || user_settings[:auth_required] || false
+
     %{verbose: false}
-    |> negotiate_auth(server_settings, user_settings)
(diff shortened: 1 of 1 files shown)


mmmries (approved):
I was reading over this PR again and I think I like your change as it is. It might make most sense to follow the convention of the go code where the client always sends authentication when it has been configured, but that would be a change in the existing behavior, whereas this PR introduces a new flag, so we can safely release this as a minor version release.

What do you think? Should we merge this as-it-is (maybe I'll add a few lines of documentation around it), or should we follow the convention in the go client and always send any auth parameters that have been configured by the client?

PR #137, opened 2022-10-15
No responders

### LINKED ISSUE #136: Quickly return for `no responders`
It looks like the only result code from making requests on a topic is `:timeout`. Does this apply to when there are no responders (which NATS immediately replies with)? If not, could we get some way to indicate the difference between when we waited the full timeout period versus when we waited no time due to lack of responders?

mmmries:
Resolves #136 

The NATS server has the ability to send reply messages with a status of `503` when a message is sent to a topic with no subscribers. This is really handy for request-reply semantics so you can get an immediate error response rather than waiting for the `receive_timeout` to expire. Because this behavior is backwards incompatible, it must be opted into by the client.

I've written this PR assuming that we also want our users to opt into this behavior rather than making it the default, but I'm open to feedback on this point. Maybe we should default the `no_responders: true` after the next major version release?

If headers are enabled, and the user has passed in `no_responders: true` with their connection options, then the server will send back HMSG payloads with an empty body and a "503" status. In order to watch for this case in the code I also added parsing of the `status` and `description` from headers. Previously these have been ignored because I didn't know of a use-case for passing them back to the user, but I think it does make sense to provide this back to the user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,9 @@
-[![circleci.com](https://circleci.com/gh/nats-io/nats.ex/tree/master.svg?style=svg)](https://circleci.com/gh/nats-io/nats.ex/tree/master)
 [![hex.pm](https://img.shields.io/hexpm/v/gnat.svg)](https://hex.pm/packages/gnat)
 [![hex.pm](https://img.shields.io/hexpm/dt/gnat.svg)](https://hex.pm/packages/gnat)
 [![hex.pm](https://img.shields.io/hexpm/l/gnat.svg)](https://hex.pm/packages/gnat)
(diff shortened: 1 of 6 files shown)


mmmries:
@autodidaddict does this API make sense to you?

autodidaddict:
To see if I've understood the API right, there's a new configuration option available in the connection settings when you start, called `no_responders`. By default it's disabled but if you enable it, then in addition to getting `:timeout` as a possible error from requests, you can also get an immediate `:no_responders` ?

If that's the case then it makes total sense

mmmries:
Yes, that's exactly how the API is working. I'll add some docs (mind if I steal some of your description above?) and then we can release this as a minor version since it's totally opt-in

PR #141, opened 2023-10-09
Adding initial services API support

autodidaddict:
* Adds support for creating microservices - developer experience is nearly identical to regular `Gnat.Server`
  * automatically adds responder for ping, stats, and info
* Fills out a few of the sparse areas of documentation
* Fixes some broken links in the docs
* Adds service request and service error to telemetry 

It's been a while since I've written some Elixir so apologies in advance if I've dumped a bunch of non-idiomatic code in this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,8 @@ For long-lived subscriptions consider using `Gnat.ConsumerSupervisor` .
 This can also be added to your supervision tree and use a supervised connection to re-establish a subscription.
 It also handles details like handling each message in a supervised process so you isolate failures and get OTP logs when an unexpected error occurs.
 
-## Microservices
-If you supply a module that implements the `Gnat.Services.Server` behavior and the `microservice` 
+## Services
+If you supply a module that implements the `Gnat.Services.Server` behavior and the `service_definition` 
 configuration field to a `Gnat.ConsumerSupervisor`, then this client will automatically take care
 of exposing the service to discovery, responding to pings, and maintaining and exposing statistics like request and error counts, and processing times.
 
--- a/README.md
+++ b/README.md
@@ -36,19 +36,19 @@ end
 # with token
 {:ok, gnat} = Gnat.start_link(%{host: '127.0.0.1', port: 4222, token: "secret", auth_required: true})
 
-# with nkeys
+# with an nkey seed
 {:ok, gnat} = Gnat.start_link(%{host: '127.0.0.1', port: 4222, nkey_seed: "SUAM...", auth_required: true})
 
-# with user credentials
+# with decentralized user credentials (JWT)
 {:ok, gnat} = Gnat.start_link(%{host: '127.0.0.1', port: 4222, nkey_seed: "SUAM...", jwt: "eyJ0eX...", auth_required: true})
 
-# connect to NGS
+# connect to NGS with JWT
 {:ok, gnat} = Gnat.start_link(%{host: "connect.ngs.global", tls: true, jwt: "ey...", nkey_seed: "SUAM..."})
 ```
 
 ## TLS Connections
 
-[Nats Server](https://github.com/nats-io/nats-server) is often configured to accept or require TLS connections.
+[NATS Server](https://github.com/nats-io/nats-server) is often configured to accept or require TLS connections.
 In order to connect to these clusters you'll want to pass some extra TLS settings to your `Gnat` connection.
 
 ``` elixir
@@ -68,6 +68,11 @@ For long-lived subscriptions consider using `Gnat.ConsumerSupervisor` .
 This can also be added to your supervision tree and use a supervised connection to re-establish a subscription.
 It also handles details like handling each message in a supervised process so you isolate failures and get OTP logs when an unexpected error occurs.
 
+## Microservices
+If you supply a module that implements the `Gnat.Services.Server` behavior and the `microservice` 
+configuration field to a `Gnat.ConsumerSupervisor`, then this client will automatically take care
+of exposing the service to discovery, responding to pings, and maintaining and exposing statistics like request and error counts, and processing times.
+
 ## Instrumentation
 
 Gnat uses [telemetry](https://hex.pm/packages/telemetry) to make instrumentation data available to clients.
@@ -85,7 +90,9 @@ iex(2)> names = [[:gnat, :pub], [:gnat, :sub], [:gnat, :message_received], [:gna
   [:gnat, :sub],
   [:gnat, :message_received],
   [:gnat, :request],
-  [:gnat, :unsub]
+  [:gnat, :unsub],
+  [:gnat, :service_request],
+  [:gnat, :service_error]
 ]
 iex(3)> :telemetry.attach_many("my listener", names, metrics_function, %{my_config: true})
 :ok
@@ -101,7 +108,8 @@ iex(6)> Gnat.pub(gnat, "topic", "ohai")
 ```
 
 The `pub` , `sub` , `request` , and `unsub` events all report the latency of those respective calls.
-The `message_received` event reports a number of messages like `%{count: 1}` because there isn't a good latency metric to report.
+The `message_received` event reports a number of messages like `%{count: 1}` because there isn't a good latency metric to report. Any microservices managed by a consumer supervisor will also report `service_request` and `service_error`. In addition to the `:topic` metadata, microservices will also include `:endpoint` and `:group` (which can be `nil`) in their telemetry reports.
+
 All of the events (except `unsub` ) include metadata with a `:topic` key so you can split your metrics by topic.
 
 ## Benchmarks
@@ -122,7 +130,8 @@ As of this commit my 2018 MacBook pro shows.
 
 ## Development
 
-Before running the tests make sure you have a locally running copy of `nats-server` ( `brew install nats-server` ).
+Before running the tests make sure you have a locally running copy of `nats-server` ([installation instructions](https://docs.nats.io/running-a-nats-service/introduction/installation)).
+
 We currently use version `2.6.6` in CI, but anything higher than `2.2.0` should be fine.
 Versions from `0.9.6` up to `2.2.0` should work fine for everything except header support.
 The typical `mix test` will run all the basic unit tests.
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -379,9 +379,7 @@ defmodule Gnat do
   @impl GenServer
   def init(connection_settings) do
     connection_settings = Map.merge(@default_connection_settings, connection_settings)
-    if :ets.whereis(:endpoint_stats) == :undefined do
-      :ets.new(:endpoint_stats, [:public, :set, :named_table])
-    end
+
     case Gnat.Handshake.connect(connection_settings) do
       {:ok, socket, server_info} ->
         parser = Parsec.new
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -3,6 +3,9 @@
 #  :waiting_for_message => receive MSG... -> :waiting_for_message
 
 defmodule Gnat do
+  @moduledoc """
+  The primary interface for interacting with NATS
+  """
   use GenServer
   require Logger
   alias Gnat.{Command, Parsec}
@@ -10,17 +13,26 @@ defmodule Gnat do
   @type t :: GenServer.server()
   @type headers :: [{binary(), iodata()}]
 
-  # A message received from NATS will be delivered to your process in this form.
-  # Please note that the `:reply_to` and `:headers` keys are optional.
-  # They will only be present if the message was received from the NATS server with
-  # headers or a reply_to topic
+  @typedoc """
+  A message received from NATS will be delivered to your process in this form.
+  Please note that the `:reply_to` and `:headers` keys are optional.
+  They will only be present if the message was received from the NATS server with
+  headers or a `reply_to` topic
+
+  * `gnat` - The Gnat connection
+  * `topic` - The topic on which the message arrived
+  * `body` - The raw payload of the message
+  * `sid` - The subscription ID corresponding to this message. You generally won't need to use this value directly.
+  * `reply_to` - A topic supplied for expected replies
+  * `headers` - A set of NATS message headers on the message
+  """
   @type message :: %{
-    gnat: t(),
-    topic: String.t(),
-    body: String.t(),
-    sid: non_neg_integer(),
-    reply_to: String.t(),
-    headers: headers()
+    required(:gnat) => t(),
+    required(:topic) => binary(),
+    required(:body) => iodata(),
+    required(:sid) => non_neg_integer(),
+    optional(:reply_to) => binary(),
+    optional(:headers) => headers()
   }
   @type sent_message :: {:msg, message()}
 
@@ -185,7 +197,7 @@ defmodule Gnat do
   ```
 
   If you want to provide a reply address to receive a response you can pass it as an option.
-  [See request-response pattern](http://nats.io/documentation/concepts/nats-req-rep/).
+  [See request-reply pattern](https://docs.nats.io/nats-concepts/core-nats/reqreply).
 
   ```
   {:ok, gnat} = Gnat.start_link()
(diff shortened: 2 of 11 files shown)


autodidaddict:
@mmmries looks like the warning message that I'm emitting regarding NATS microservices being experimental is messing up the build, but I'm not sure why.

mmmries:
Hey @autodidaddict I think the build error you're seeing is because we do a `compile --warnings-as-errors`. This is mostly a sanity check that we aren't adding code to the project that raises errors with the compiler, but it sounds like you were specifically wanting to add an error for users of the library. So we might want to drop that `--warnings-as-errors` from the CI job for now?

autodidaddict:
> Hey @autodidaddict I think the build error you're seeing is because we do a `compile --warnings-as-errors`. This is mostly a sanity check that we aren't adding code to the project that raises errors with the compiler, but it sounds like you were specifically wanting to add an error for users of the library. So we might want to drop that `--warnings-as-errors` from the CI job for now?

The builds are passing now. Apparently I'd been using some language features that didn't exist in much older versions of Elixir, so now the CI is using 1.13 and up, which feels pretty safe.

mmmries:
Screenshot 2023-10-12 at 10 54 40

I'm thinking about this some more and dropping support for any elixir versions before 1.13 means that only the last 2 years of releases would be supported. I think most of the community tends to adopt new releases pretty quickly, but I'm imagining some poor soul on an engineering team who needs to update dependencies to get a bugfix or something and now they have to update elixir/otp versions. Do you know which syntactical things required the newer versions? I'm wondering if the code is a lot worse if we change to be compatible back to say Elixir 1.10?

autodidaddict:
I think there was a map function that required 1.13. Also when I was trying to get things to work I got some messages about dependencies requiring 1.13. I can poke around again and see, though

autodidaddict:
@mmmries ok looks like we can go back as far as 1.10 now

README.md
```diff
@@ -122,7 +130,8 @@ As of this commit my 2018 MacBook pro shows.
 
 ## Development
 
-Before running the tests make sure you have a locally running copy of `nats-server` ( `brew install nats-server` ).
+Before running the tests make sure you have a locally running copy of `nats-server` ([installation instructions](https://docs.nats.io/running-a-nats-service/introduction/installation)).
```
mmmries, line 133: 👍

lib/gnat/consumer_supervisor.ex
```diff
@@ -148,4 +188,33 @@ defmodule Gnat.ConsumerSupervisor do
...
+
+    subscriptions = Enum.map(endpoints, fn(ep) ->
+      subject = ServiceResponder.derive_subscription_subject(ep)
+      queue_group = Map.get(ep, :queue_group, @default_service_queue_group)
+      {:ok, subscription} = Gnat.sub(connection_pid, self(), subject, queue_group: queue_group)
+
+      subscription
+    end)
+
+    %{state | svc_responder_pid: responder_pid, subscriptions: subscriptions }
```
mmmries, line 217: I'm part way through reviewing this and I have to jump on a work call soon. I'm trying to wrap my head around the use-case for the `ets` table and `responder_pid`. But I wanted to provide some of the early feedback before my work day gets busy

autodidaddict: Looking forward to your feedback. The `responder_pid` is to deal with the fact that whenever a consumer supervisor is managing a service, it creates a new child (`ServiceResponder`) that is responsible for handling the protocol requests like `PING`, `STATS`, and `INFO`. This responder maintains this automatically and uses the ets table as a way to allow both the `Gnat.Services.Server` and the responder to access the stat counters for processing time, errors, and success

lib/gnat.ex
```diff
@@ -367,6 +379,9 @@ defmodule Gnat do
   @impl GenServer
   def init(connection_settings) do
     connection_settings = Map.merge(@default_connection_settings, connection_settings)
+    if :ets.whereis(:endpoint_stats) == :undefined do
```
mmmries, line 382: Do you think we should setup the ETS table in the `ConsumerSupervisor` when we are doing the `initialize_as_service` call? That way we don't create the ETS table for people who aren't using services?

Maybe setting up an empty ETS table doesn't really matter? I don't think it's much overhead

autodidaddict: Yeah we could probably move it. As for overhead, I think an empty ETS table is just a `ref` and shouldn't take much.

autodidaddict: Pushed the refactor

lib/gnat/services/service_responder.ex
```diff
@@ -0,0 +1,317 @@
...
+
+  @spec start_link(map(), keyword()) :: GenServer.on_start
+  def start_link(settings, options \\ []) do
+    GenServer.start_link(__MODULE__, settings, options)
+  end
+
+  @impl GenServer
+  @spec init(map) :: {:ok, state()} | {:stop, String.t}
+  def init(settings) do
+    Process.flag(:trap_exit, true)
```
mmmries, line 38: I'm not seeing where we handle exit messages down below. Shouldn't we have a `handle_info` clause to accept exit messages? Or are we capturing exist just to get the `terminate` notification to try to unsubscribe?

lib/gnat/services/service_responder.ex
```diff
@@ -0,0 +1,317 @@
+defmodule Gnat.Services.ServiceResponder do
```
mmmries, line 1: In this project overall I've been writing GenServers with a structure of:

1. Public Interface Functions
2. GenServer Callbacks
3. Private functions

I think Dave Thomas was the first person I heard this from, but I've seen it around in the community. Can we organize this module in a similar way?

lib/gnat/services/service_responder.ex
```diff
@@ -0,0 +1,317 @@
...
+          }
+        end)
+      }
+
+      Gnat.pub(gnat, rt, Jason.encode!(output))
+    end
+  end
+
+  @spec should_respond(list, String.t, String.t) :: boolean()
+  defp should_respond(tail, service_name, instance_id) do
```
mmmries, line 221: In your other predicate functions down below like `valid_version?` you have the question mark on the end. Do you think it would make sense to add here as well for consistency?

lib/gnat/services/service_responder.ex
```diff
@@ -0,0 +1,317 @@
+defmodule Gnat.Services.ServiceResponder do
+  @moduledoc false
+
+  require Logger
+  alias Gnat.Services.WireProtocol
+  @subscription_subject "$SRV.>"
```
mmmries, line 6: Should this be a dynamically calculation subject? it seems like the convention is to append the service name to the `$SRV` prefix? If our Responder subscribes to everything on this prefix, will it receive PING, STATS etc for other services?

test/test_helper.exs
```diff
@@ -32,6 +32,30 @@ defmodule RpcEndpoint do
 end
 spawn(&RpcEndpoint.init/0)
 
+defmodule ExampleService do
+  use Gnat.Services.Server
+
+  def request(%{topic: "calc.add", body: body}, _endpoint, _group) when body == "foo" do
+    :timer.sleep(100)
```
mmmries, line 39: it looks like we are just using this delay to validate that we have a non-zero processing time. How about bumping it down to 10ms? Avoid slowing down tests, but still provide a non-zero latency?

lib/gnat/services/server.ex
```diff
@@ -0,0 +1,167 @@
...
+    optional(:metadata) => map()
+  }
+
+
+  @doc false
+  def execute(module, message, responder_pid) do
+    try do
+      {endpoint, group} = ServiceResponder.lookup_endpoint(responder_pid, message.topic)
+
+      case :timer.tc(fn -> apply(module, :request, [message, endpoint, group]) end) do
```
mmmries, line 118: I'm pretty sure `:timer.tc` provides an elapsed number of microseconds, not nanoseconds?

mmmries: yeah it looks like the default is microseconds, but you an pass a time-unit to get it back as nanoseconds if you want https://www.erlang.org/doc/man/timer#tc-1

mmmries (commented):
Left lots of smaller comments. One thing I'm wondering about overall is if we need the Responder as a separate process? Would it be possible to just subscribe to some extra topics from the ConsumerSupervisor? And when we see messages come in with the `$SRV` prefix we just do the counter logic in a task?

I don't understand everything yet about the `lookup_endpoint` flow, so this may not be possible, but I think it could be pretty simple if we just intercepted the protocol messages and handled those separately.

lib/gnat/services/service_responder.ex
autodidaddict: Excellent point. This should've had a ?

lib/gnat/services/service_responder.ex
autodidaddict: All services must respond to the action types without the service name. The `PING` and `INFO` commands are used to interrogate which services are available, so the responder needs to be listening on `$SRV.PING` and `$SRV.INFO` etc. There's a check below to see that if a service name is supplied, it matches the one defined in the responder.

autodidaddict:
> Left lots of smaller comments. One thing I'm wondering about overall is if we need the Responder as a separate process? Would it be possible to just subscribe to some extra topics from the ConsumerSupervisor? And when we see messages come in with the `$SRV` prefix we just do the counter logic in a task?
> 

I originally started out with this design but then moved to the responder. I didn't want to add a bunch of overhead and conditionals to the consumer supervisor that would only be meaningful when it's supervising a service. I'm open to the idea of going back to that if it can be clean with no overhead.

> I don't understand everything yet about the `lookup_endpoint` flow, so this may not be possible, but I think it could be pretty simple if we just intercepted the protocol messages and handled those separately.

This is done so that the `Gnat.Services.Server` callback can include the endpoint name in the callback. This is to give it parity with other client libraries that allow a service's handler to capture the endpoint name.

lib/gnat/services/service_responder.ex
autodidaddict: It's basically just to get the terminate. If it's better to handle the exit infos as well, I'll see if I can figure that out :)

autodidaddict:
@mmmries unsure why the build is failing this time. Running `:timer.tc(fn -> foo end, :nanosecond)` works just fine in iex. 

I'm almost inclined to just get microsecond and multiply it by 1000. I can't imagine why we'd need nanosecond level precision on general-purpose service execution stats.

autodidaddict:
@mmmries ok I put the timer command back to microseconds and am just multiplying to get nanos. It looks like this function has changed across multiple OTP versions and so that might explain why it works locally but fails the build. 

Everything is building now

mmmries (approved):
Sorry I've been slow to review here. I'm good with moving this forward and just left a few comments on things we could optimize for performance/simplicity. But I think they can all be addressed in future iterations.

lib/gnat/services/service_responder.ex
```diff
@@ -0,0 +1,319 @@
...
+  @op_ping "PING"
+  @op_stats "STATS"
+  @op_info "INFO"
+
+  @idx_requests 1
+  @idx_errors 2
+  @idx_processing_time 3
+
+  @name_regex ~r/^[a-zA-Z0-9_-]+$/
+  @version_regex ~r/^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/
```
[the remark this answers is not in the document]
autodidaddict: This comes from the semver.org website. The NATS services protocol dictates that the version string must be valid semver, so a multi-line string would be rejected outright.

This is the same regex that the Go client uses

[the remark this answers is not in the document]
mmmries: 👍 Thanks for the details. That definitely makes me feel better about including it.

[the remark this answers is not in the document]
autodidaddict: yeah I probably should've left a comment above that indicating where I got that regex from.

mmmries:
I think this is ready to be merged and released. My typical workflow for this has been:

* Merge the PR
* Pull the latest `main` branch locally
* Modify the `mix.exs` file to change the version number, and edit the `CHANGELOG.md` with a few comments
* Run `mix hex.publish` and make sure the new version shows up in hexpm
* Commit the local changes and tag the commit with `v1.7.0` and push the commit and tag straight to the repo

Feel free to modify those steps if you have a workflow that works better for you, that's just what I've done in the past. Do you want to merge and release this change?

lib/gnat/services/server.ex
```diff
@@ -0,0 +1,167 @@
...
+  * `queue_group` - An optional queue group for service subscriptions. If left off, "q" will be used.
+  * `description` - An optional description of the service
+  * `metadata` - An optional string->string map of service metadata
+  * `endpoints` - A required list of service endpoints. All services must have at least one endpoint
+  """
+  @type service_configuration :: %{
+    required(:name) =>         binary(),
+    required(:version) =>      binary(),
+    required(:endpoints) =>    [endpoint_configuration()],
+    optional(:queue_group) => binary(),
```
mmmries, line 89: @autodidaddict I didn't notice this during the first PR, but I'm not sure what the purpose of this service-level queue group would be? It looks like when it's specific you use it in the `Gnat.sub` call, but when I looked at the ADR document, it seems like queue groups only apply to endpoints, not to services?

https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-32.md

I think specifying a queue group for the `$SRV.>` subscription would be an error since we want all instances of the service to decide whether they should respond to a particular services call?

lib/gnat/services/service_responder.ex
```diff
@@ -0,0 +1,319 @@
...
+  end
+
+  @impl true
+  def handle_cast({:record_request, subject, elapsed_ns}, state) do
+    counters = case :ets.lookup(:endpoint_stats, subject) do
+      [{^subject, counters, _last_error}] ->
+        counters
+      [] ->
+        c = :counters.new(3, [:atomics])
+        :ets.insert(:endpoint_stats, {subject, c, nil})
```
mmmries, line 144: I'm also not sure why we were inserting ETS records for the requests and errors? It looks like we only ever provide stats from the counters. Maybe in an earlier revision you were using ETS to count the requests and errors?

mmmries: Oh, maybe it's related to the `last_error` part of the stats callback? I didn't notice that on my first read-through of the spec, but the `last_error` and `data` keys seem a bit strange...

autodidaddict: the second element in the tuple is the counter set that contains the request and error counts.

lib/gnat/services/server.ex
autodidaddict: This is a good catch. We shouldn't be queue-subscribing on `$SRV` at all.

lib/gnat/services/service_responder.ex
autodidaddict: The only reason it's in `ets` is to allow a lookup of the counter set by subject when a request comes in. This lets us pull out the counter reference, increment, and move on.

mmmries:
I'm going to take a stab at a PR where we avoid having to callback to the ServiceResponder to get endpoint details. I'm not exactly sure how to describe what I'm thinking, so I'll just write it up in code and open as a PR to get your thoughts

autodidaddict:
@mmmries as you suspected, there was some scar tissue left over from having 2 different designs in my head at once. I've addressed the concerns in #144

autodidaddict:
> I'm going to take a stab at a PR where we avoid having to callback to the ServiceResponder to get endpoint details. I'm not exactly sure how to describe what I'm thinking, so I'll just write it up in code and open as a PR to get your thoughts

Might want to take a look at the newest PR #144 to make sure you don't re-invent wheels

Issue #142, opened 2023-10-10
Gnat.ConnectionSupervisor bug with erlang OTP 26


Qizot:
# Problem
In OTP 26 the genserver's behavior changed when it comes to returning `{:stop, reason}` from an `init` callback. The process currently exits with reason normal which is not triggering the `EXIT` signal from `start_link` function.

Currently the `Gnat.ConnectionSupervisor` relies on this behaviour. It expectes to receive an `EXIT` signal and then schedules a connection attempt. But when it doesn't receive the signal (OTP 26) it just does nothing and will never retry restarting Gnat connection.

## Solution
Besides relying on exit signal, the supervisor should reschedule connection attempt if `Gnat.start_link` returns an error tuple.

Example implementation:
```elixir
defmodule Gnat.ConnectionSupervisor do
  use GenServer

  require Logger

  @type settings_t :: %{
          connection_name: atom(),
          connection_settings: Gnat.connection_settings()
        }

  @spec start_link(map(), keyword()) :: GenServer.on_start()
  def start_link(settings, options \\ []) do
    GenServer.start_link(__MODULE__, settings, options)
  end

  @impl true
  def init(opts) do
    Process.flag(:trap_exit, true)

    state = %{
      connection: nil,
      connection_settings: opts.connection_settings,
      connection_name: opts.connection_name,
      connect_attempt: 0
    }

    {:ok, state, {:continue, :attempt_connect}}
  end

  @impl true
  def handle_continue(:attempt_connect, state) do
    case Gnat.start_link(state.connection_settings, name: state.connection_name) do
      {:ok, connection} ->
        Logger.info("NATS connection established")
        {:noreply, %{state | connection: connection, connect_attempt: 0}}

      {:error, reason} ->
        Logger.error("Failed to establish NATS connection due to: #{inspect(reason)}")

        {:noreply, %{state | connection: nil, connect_attempt: state.connect_attempt + 1},
         {:continue, :schedule_reconnect}}
    end
  end

  @impl true
  def handle_continue(:schedule_reconnect, state) do
    backoff = backoff_time(state)

    Logger.info("Retrying to connect in #{backoff}ms")

    Process.send_after(self(), :try_connect, backoff)

    {:noreply, state}
  end

  @impl true
  def handle_info(:try_connect, state) do
    {:noreply, state, {:continue, :attempt_connect}}
  end

  def handle_info({:EXIT, pid, reason}, %{connection: pid} = state) do
[...]

PR #143, opened 2023-10-12
provide connection retries in OTP 26

mmmries:
This addresses #142 

I tested this locally on both OTP 25 and OTP 26. I could reproduce the reported issue with the current `main` branch, but this new version of the code worked on both OTP 25 and OTP 26. I tried to do a minimal amount of changes here. I didn't want to build an exponential backoff for instance, I just wanted to fix the bug and keep the behavior unchanged.

### DIFF
--- a/lib/gnat/connection_supervisor.ex
+++ b/lib/gnat/connection_supervisor.ex
@@ -62,12 +62,15 @@ defmodule Gnat.ConnectionSupervisor do
       {:ok, gnat} -> {:noreply, %{state | gnat: gnat}}
       {:error, err} ->
         Logger.error "failed to connect #{inspect err}"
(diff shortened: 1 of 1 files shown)


mmmries:
@autodidaddict I'm thinking that we should get this in and get it released pretty quickly since it's confirmed to break the `ConnectionSupervisor` on OTP 26?

autodidaddict (approved):

autodidaddict:
> @autodidaddict I'm thinking that we should get this in and get it released pretty quickly since it's confirmed to break the `ConnectionSupervisor` on OTP 26?

We can probably include this in the release that will contain the services support?

mmmries:
👍 that sounds good to me

PR #144, opened 2023-10-21
Fixes unnecessary use of ETS and misleading docs re: queue_group

autodidaddict:
Good catch on finding these @mmmries .

PR #145, opened 2023-10-21
A draft of simplifying the service implementation

mmmries:
After reviewing #141 a bit further and reviewing the ADR document, I had a few thoughts about how to simplify and optimize our service implementation:

* Normalize the user-provided service config into a known structure with guaranteed keys
  * this reduces the need for conditional code later on
* Allocate `:counters` during initialization and skip the use of ETS
* Isolate as much of the knowledge about structure of the `service` datastructure to one module as possible. Use functions to pull out what is needed for service responder etc.
* Avoid starting ServiceResponder as an additional process, just add an additional subscription to the `ConsumerSupervisor`

Note: I wrote this before #144 was opened, but it looks like we were thinking of some of these same things


bench/service_bench.exs
```diff
@@ -0,0 +1,56 @@
+defmodule EchoService do
```
mmmries, line 1: I added a benchmark to try to measure the impact of these changes, but it turns out this branch only make a ~5% improvement in throughput.

We can get around 13.7k service messages per second on both the main branch and about 14.5k service messages per second on this branch. 🤷

lib/gnat/consumer_supervisor.ex
```diff
@@ -84,23 +79,15 @@ defmodule Gnat.ConsumerSupervisor do
...
-        {:ok, state}
-
-      Map.has_key?(settings, :consuming_function) ->
-        state = Map.put(state, :consuming_function, settings.consuming_function)
-        send self(), :connect
-        {:ok, state}
-
-      true ->
-        {:error, "You must provide a module or consuming function for the consumer supervisor"}
+    with {:ok, state} <- maybe_append_service(state, settings),
```
mmmries, line 85: Curious if you'll like this change. The `cond` block was a bit tricky to read and I wanted to validate the user-input about the service definition during process startup rather than crashing the process later when the connection became available.

So I used a `with` block here to show the happy path, and I perform all of the service configuration validation right here (as well as initializing service counters etc). This way our stats will persist across re-connects and if the user has an invalid validation they'll get a helpful error at boot time rather than a process crash later

lib/gnat/consumer_supervisor.ex
```diff
@@ -189,35 +174,65 @@ defmodule Gnat.ConsumerSupervisor do
     end
   end
 
-  defp initialize_as_manual_consumer(state, connection_pid) do
-    subscriptions = Enum.map(state.subscription_topics, fn(topic_and_queue_group) ->
+  defp subscribe_to_topics(%{service: service}, connection_pid) do
```
mmmries, line 177: I generally prefer to do pattern matching in my function heads rather than `if` checks inside the function bodies. I don't feel strongly about this preference, but I like how it looks in this case. What do you think?

lib/gnat/services/service_responder.ex
```diff
@@ -2,257 +2,40 @@ defmodule Gnat.Services.ServiceResponder do
...
-  @impl true
-  def handle_call({:lookup_endpoint, subject}, _from, state) do
-    res = case Map.get(state.subject_map, subject) do
-      nil ->
-        {nil, nil}
-      {epname, groupname} ->
-        {epname, groupname}
+  def maybe_respond(%{topic: topic} = message, service) do
+    case String.split(topic, ".") do
+      ["$SRV", @op_ping | rest] -> handle_ping(rest, service, message)
```
mmmries, line 13: I liked pattern matching to grab the prefix, op and `rest` or `tail` this way rather than using `List.slice`. What do you think?

test/gnat/consumer_supervisor_test.exs
```diff
@@ -28,7 +31,40 @@ defmodule Gnat.ConsumerSupervisorTest do
     assert {:ok, %{body: "500 error"}} = Gnat.request(:test_connection, "calc.sub", "error")
   end
 
-  test "service endpoint counters working" do
+  test "service endpoint ping response" do
```
mmmries, line 34: added tests for the PING and INFO service calls as well

autodidaddict (approved):
I don't know if there's anything that you want to incorporate from #144 , but this all looks better than what I had. It's been too long since I've written idiomatic Elixir.

lib/gnat/services/service_responder.ex
autodidaddict: yeah this looks much cleaner

mmmries:
I pulled in some of the typedoc and function doc improvements from #144 into this PR. Thanks for reviewing and sorry we ended tackling some of the same work. But I was glad to see how our thinking was converging.

I'm on vacation so I suddenly had extra time for this project. I'll try to follow your lead on creating a github release

PR #146, opened 2023-11-29
Incorporing the existing Jetstream code into the main Gnat client

autodidaddict:
this adds the JetStream code with no refactoring (other than module names) from the other Jetstream repository. The purpose is to make the JetStream code more discoverable and accessible to people who find Gnat and are looking for stream, key-value and object store functionality.

if we want to refactor the APIs, we can do that in subsequent PRs.

This also bumps to 1.8.0 as per semver guidelines.


autodidaddict:
@mmmries still trying to figure out why dialyzer is rejecting this one set of code.

autodidaddict:
Should we maybe just pull out the broadway thing to get dialyzer to be happy and then add it back in later?

autodidaddict:
Going to merge this into main so I can try and get the GH action running

Issue #152, opened 2024-04-09
Cannot list KV buckets


lsxliron:
Missing functionality to list existing buckets.

Equivalent CLI command would be `nats kv ls`

mmmries:
This was addressed in #153

PR #153, opened 2024-04-09
Added a function to list available buckets

lsxliron:
Addresses #152 

```elixir
{:ok, conn} = Gnat.start_link()
# {:ok, #PID<0.301.0>}
Gnat.Jetstream.API.KV.list_buckets(conn)
# {:ok, ["bucket1", "bucket2", "bucket3"]}
```

### DIFF
--- a/lib/gnat/jetstream/api/kv.ex
+++ b/lib/gnat/jetstream/api/kv.ex
@@ -265,6 +265,28 @@ defmodule Gnat.Jetstream.API.KV do
     end
   end
 
+  @doc """
+  Returns a list of all the buckets in the KV
+  """
+  @spec list_buckets(conn :: Gnat.t()) :: {:error, term()} | {:ok, list(String.t())}
+  def list_buckets(conn) do
+    case Stream.list(conn) do
+      {:ok, %{streams: streams}} ->
+        {:ok,
+         streams
+         |> Enum.flat_map(fn bucket ->
+           case Stream.is_kv_bucket_stream(bucket) do
+             true -> [bucket |> String.trim_leading(@stream_prefix)]
+             _ -> []
+           end
+         end)}
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+
   @doc false
   def stream_name(bucket_name) do
     "#{@stream_prefix}#{bucket_name}"
--- a/lib/gnat/jetstream/api/kv.ex
+++ b/lib/gnat/jetstream/api/kv.ex
@@ -266,6 +266,17 @@ defmodule Gnat.Jetstream.API.KV do
   end
 
   @doc """
+  Returns true if the provided stream is a KV bucket, false otherwise
+
+  ## Parameters
+  * `stream_name` - the stream name to test
+  """
+  @spec is_kv_bucket_stream?(stream_name :: binary()) :: boolean()
+  def is_kv_bucket_stream?(stream_name) do
+    String.starts_with?(stream_name, "KV_")
+  end
+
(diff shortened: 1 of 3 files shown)


mmmries:
Thanks @lsxliron for opening this PR! I have a couple of stylistic suggestions here, but no concerns about the functionality or general approach. Would you prefer if i just merged this and just made a few of my own tweaks before releasing, or add my comments inline on the PR and let you make the changes?

lsxliron:
Sure thing @mmmries. Please feel free to share your suggestions. I can make changes if needed.

lib/gnat/jetstream/api/kv.ex
```diff
@@ -265,6 +265,28 @@ defmodule Gnat.Jetstream.API.KV do
     end
   end
 
+  @doc """
+  Returns a list of all the buckets in the KV
+  """
+  @spec list_buckets(conn :: Gnat.t()) :: {:error, term()} | {:ok, list(String.t())}
+  def list_buckets(conn) do
+    case Stream.list(conn) do
```
mmmries, line 273: I would swap this `case` statement for a `with` since it has an obvious happy-path pattern.

`with {:ok, %{streams: streams}} <- Stream.list(conn) do`

lib/gnat/jetstream/api/kv.ex
```diff
@@ -265,6 +265,28 @@ defmodule Gnat.Jetstream.API.KV do
...
+      {:ok, %{streams: streams}} ->
+        {:ok,
+         streams
```
mmmries, line 276: I would usually build a `stream_names = Enum.flat_map(...)` on it's own before assembling the `{:ok, stream_names}` tuple. That avoids extra indentation and is easier to read the flow of the function. This is a subjective  opinion for sure, but most of the codebase in this project follows this pattern, so I think it's worth applying here.

lib/gnat/jetstream/api/stream.ex
```diff
@@ -441,6 +441,17 @@ defmodule Gnat.Jetstream.API.Stream do
...
   end
 
+  @doc """
+  Returns true if the provided stream is a KV bucket, false otherwise
+
+  ## Parameters
+  * `stream_name` - the stream name to test
+  """
+  @spec is_kv_bucket_stream(stream_name :: binary()) :: boolean()
+  def is_kv_bucket_stream(stream_name) do
```
mmmries, line 451: I would suggest moving this function to the `KV` module. So far I've been keeping all of the KV-specific knowledge in that module and similarly keeping everything related to `Object` in its own module, so the stream module really just has knowledge of general streams.

I'm definitely open to feedback here if you had a specific reason for putting it into the stream module.

mmmries (changes requested):
I also noticed that the main branch of elixir is showing a bunch of compiler warnings and the latest version of nats server is breaking on some of our unit tests. I'll open a separate PR to address those items, so don't worry about those CI checks here.

lib/gnat/jetstream/api/stream.ex
lsxliron: No specific reason other than the fact that it's dealing with the actual stream.
Can definitely be moved to the KV module

mmmries:
Thanks for the great contribution and fast changes @lsxliron ❤️💙💜💚💛🧡🤎🖤 

I'm going to try to sort out why we're getting different errors on some CI runs with more recent versions of Elixir/NATS and if I can sort that out within a day or two, I'll include any of those fixes in the next release along with your change

PR #157, opened 2024-04-19
Emitting `:key_deleted` when key is purged

### LINKED ISSUE #156: KV watcher does not fire when key is purged
### Current Behavior
watching a bucket using [Gnat.Jetstream.API.KV.watch/3](https://hexdocs.pm/gnat/Gnat.Jetstream.API.KV.html#watch/3) does not fire an event when a key is purged

```elixir
# create a bucket
bucket = "my_bucket"

# connect
{:ok, pid} = Gnat.start_link()

# create a bucket
{:ok, _info} = Gnat.Jetstream.API.KV.create_bucket(pid, bucket, history: 10)


handler = fn action, key, value ->
  IO.puts("ACTION --> #{inspect(action)}")
  IO.puts("KEY --> #{inspect(key)}")
  IO.puts("VALUE --> #{inspect(value)}")
end

# start watching the bucket
{:ok, _pid} = Gnat.Jetstream.API.KV.watch(pid, bucket, handler)

# Put the first key
:ok = Gnat.Jetstream.API.KV.create_key(pid, bucket, "my_key_1", "my_value_1")
# output:
#   ACTION --> :key_added
#   KEY --> "my_key_1"
#   VALUE --> "my_value_1"

# delete the first key
:ok = Gnat.Jetstream.API.KV.delete_key(pid, bucket, "my_key_1")
# output:
#   ACTION --> :key_deleted
#   KEY --> "my_key_1"
#   VALUE --> ""

# put the second key
:ok = Gnat.Jetstream.API.KV.put_value(pid, bucket, "my_key_2", "my_value_2")
# output:
#   ACTION --> :key_added
#   KEY --> "my_key_2"
#   VALUE --> "my_value_2"

# purge the second key
:ok = Gnat.Jetstream.API.KV.purge_key(pid, bucket, "my_key_2")
# actual output:
#   :ok
#
# expected output:
#   ACTION --> :key_deleted
#   KEY --> "my_key_2"
#   VALUE --> ""

```
### Expected Behavior
When a key is purged the watcher should emit the same message as if the key was deleted.

lsxliron:
- Now `:key_deleted` event is emitted when wither purging or deleting a key
- Moved operations module attributes to KV module for better readability

Fixes #156

### DIFF
--- a/lib/gnat/jetstream/api/kv.ex
+++ b/lib/gnat/jetstream/api/kv.ex
@@ -257,13 +257,13 @@ defmodule Gnat.Jetstream.API.KV do
   ## Parameters
   - `headers` - a list of headers to test
 
-  ## Example
+  ## Examples
 
-      iex> is_delete_operation([{"kv-operation", "DEL"}])
+      iex> is_delete_operation?([{"kv-operation", "DEL"}])
       true
-      iex> is_delete_operation([{"kv-operation", "PURGE"}])
+      iex> is_delete_operation?([{"kv-operation", "PURGE"}])
       true
-      iex> is_delete_operation([{"kv-operation", "ADD"}])
+      iex> is_delete_operation?([{"kv-operation", "ADD"}])
       true
   """
   @spec is_delete_operation?(headers :: Gnat.headers()) :: boolean()
--- a/lib/gnat/jetstream/api/kv.ex
+++ b/lib/gnat/jetstream/api/kv.ex
@@ -264,7 +264,7 @@ defmodule Gnat.Jetstream.API.KV do
       iex> is_delete_operation?([{"kv-operation", "PURGE"}])
       true
       iex> is_delete_operation?([{"kv-operation", "ADD"}])
-      true
+      false
   """
   @spec is_delete_operation?(headers :: Gnat.headers()) :: boolean()
   def is_delete_operation?(headers) do
--- a/lib/gnat/jetstream/api/kv.ex
+++ b/lib/gnat/jetstream/api/kv.ex
@@ -9,6 +9,9 @@ defmodule Gnat.Jetstream.API.KV do
   @stream_prefix "KV_"
   @subject_prefix "$KV."
   @two_minutes_in_nanoseconds 1_200_000_000
+  @operation_header "kv-operation"
+  @operation_del "DEL"
+  @operation_purge "PURGE"
 
   @type bucket_options ::
           {:history, non_neg_integer()}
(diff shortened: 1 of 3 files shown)


mmmries (commented):
Do you think there will ever be a case where someone cares about the difference between delete vs purge? I don't use the KV functionality much, so I'm not sure about the semantics of this API

mmmries:
I don't know go very well, but I think this bit of the go-client is exposing purge events separately from delete events? https://github.com/nats-io/nats.go/blob/9d4b227179d60d6996c32b4b889d4e325ee06f78/kv.go#L982-L989

lsxliron:
@mmmries i just started using the KV so I’m not an expert. The end result is the same, the key is removed from the bucket (history stays when using del and removed when using purge). 

However, when the watcher process starts, you will still get historical events for a key that was deleted but not for a key that was purged.

lib/gnat/jetstream/api/kv.ex
```diff
@@ -248,10 +251,34 @@ defmodule Gnat.Jetstream.API.KV do
...
+      iex> is_delete_operation?([{"kv-operation", "ADD"}])
+      false
+  """
+  @spec is_delete_operation?(headers :: Gnat.headers()) :: boolean()
+  def is_delete_operation?(headers) do
+    headers
+    |> Enum.filter(fn {k, v} ->
+      k == @operation_header and (v == @operation_del or v == @operation_purge)
+    end)
+    |> length() > 0
```
davydog187, line 275:
```suggestion
    Enum.any?(headers, fn {k, v} ->
      k == @operation_header and (v == @operation_del or v == @operation_purge)
    end)
```

In practice this probably isn't a huge issue as written since the size of headers is probably small, but this is equivalent. At worst it makes a full pass once, at best it short-circuits early

lib/gnat/jetstream/api/kv.ex
```diff
@@ -248,10 +251,34 @@ defmodule Gnat.Jetstream.API.KV do
...
+
+      iex> is_delete_operation?([{"kv-operation", "DEL"}])
+      true
+      iex> is_delete_operation?([{"kv-operation", "PURGE"}])
+      true
+      iex> is_delete_operation?([{"kv-operation", "ADD"}])
+      false
+  """
+  @spec is_delete_operation?(headers :: Gnat.headers()) :: boolean()
+  def is_delete_operation?(headers) do
```
davydog187, line 270:
```suggestion
  @spec delete_operation?(headers :: Gnat.headers()) :: boolean()
  def delete_operation?(headers) do
```

As an Elixir convention, using the `?` is correct for a function that returns a boolean, but the `is_` prefix is reserved for guards. See https://hexdocs.pm/elixir/naming-conventions.html#trailing-question-mark-foo

mtmk:
apologies I didn't see this PR. #191 addresses key purges. I will close this now, but please feel free to reopen if missed something else.

PR #159, opened 2024-04-25
Page over stream

mmmries:
I did a bit of experimenting with having a functionality interface for a pull consumer that I think might be useful for people. The basic idea is that the caller calls `Pager.init(...)` and then repeatedly calls `Pager.page` to get the next "page" of messages. When the `page` function returns `{:done, messages}` it means the consumer has reached the current end of the stream and the client should then call `Pager.cleanup` to immediately cleanup the subscription and consumer.

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -25,6 +25,8 @@ defmodule Gnat do
   * `sid` - The subscription ID corresponding to this message. You generally won't need to use this value directly.
   * `reply_to` - A topic supplied for expected replies
   * `headers` - A set of NATS message headers on the message
+  * `status` - Similar to an HTTP status, this is present for messages with headers and can indicate the specific purpose of a message. Example `status: "408"`
+  * `description` - A string description of the `status`
   """
   @type message :: %{
     required(:gnat) => t(),
@@ -32,7 +34,9 @@ defmodule Gnat do
     required(:body) => iodata(),
     required(:sid) => non_neg_integer(),
     optional(:reply_to) => binary(),
-    optional(:headers) => headers()
+    optional(:headers) => headers(),
+    optional(:status) => String.t(),
+    optional(:description) => String.t()
   }
   @type sent_message :: {:msg, message()}
 
--- a/lib/gnat/jetstream/jetstream.ex
+++ b/lib/gnat/jetstream/jetstream.ex
@@ -25,7 +25,7 @@ Acknowledges the message was handled and requests delivery of the next message t
 subject. Only applies to Pull-mode.
 """
 @spec ack_next(message :: Gnat.message(), consumer_subject :: binary()) :: :ok
-def ack_next(%{reply_to: nil}, consumer_subject) do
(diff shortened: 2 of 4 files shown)


lib/gnat/jetstream/pager.ex
```diff
@@ -0,0 +1,76 @@
...
+  alias Gnat.Jetstream
+  alias Gnat.Jetstream.API.{Consumer, Util}
+
+  @opaque pager :: map()
+  @type message :: Gnat.message()
+
+  def init(conn, stream_name, opts) do
+    name = "gnat_stream_pager_#{Util.nuid()}"
+
+    first_seq = Keyword.fetch!(opts, :from_seq)
```
mmmries, line 13: I imagine that we would end up allowing other options like `since: utc_datetime` or `limit: 1000` in these opts

lib/gnat/jetstream/pager.ex
```diff
@@ -0,0 +1,75 @@
...
+    with {:ok, _config} <- Consumer.create(conn, consumer),
+         {:ok, sub} <- Gnat.sub(conn, self(), inbox) do
+      state =
+        %{
+          conn: conn,
+          stream_name: stream_name,
+          consumer_name: name,
+          domain: nil,
+          inbox: inbox,
+          batch: 10,
```
mmmries, line 37: We should also allow for `domain` and `batch` to be passed in via opts

lib/gnat/jetstream/pager.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Gnat.Jetstream.Pager do
+  @moduledoc false
```
mmmries, line 2: Keeping this undocumented until we feel comfortable with the API

lib/gnat/jetstream/jetstream.ex
```diff
@@ -10,7 +10,9 @@ Sends `AckAck` acknowledgement to the server.
 Acknowledges a message was completely handled.
 """
 @spec ack(message :: Gnat.message()) :: :ok
-def ack(message)
+def ack(%{reply_to: nil}) do
```
mmmries, line 13: I ran into a case where I got a `%{status: "408", reply_to: nil}` message at the end of the stream and my code tried to ack the message which resulted in a TCP error and it killed my gnat connection. So I thought it would be good to validate the `reply_to` here with a more useful error message

lib/gnat.ex
```diff
@@ -25,14 +25,18 @@ defmodule Gnat do
   * `sid` - The subscription ID corresponding to this message. You generally won't need to use this value directly.
   * `reply_to` - A topic supplied for expected replies
   * `headers` - A set of NATS message headers on the message
+  * `status` - Similar to an HTTP status, this is present for messages with headers and can indicate the specific purpose of a message. Example `status: "408"`
```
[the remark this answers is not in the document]
mmmries: Hmmmm I can find several examples in the ADRs, but I don't think we can provide an exhaustive list since clients can publish their own messages with headers and provide their own status codes

PR #161, opened 2024-06-19
fix: add stream fields `allow_direct` and `mirror_direct`

djcarpe:
When updating a stream that has `allow_direct` stream set to `true` will overwrite it within the server to `false`. It was a rather unexpected behavior. If there is a better way to implement this, please let me know. Thanks!

### DIFF
--- a/lib/gnat/jetstream/api/stream.ex
+++ b/lib/gnat/jetstream/api/stream.ex
@@ -11,6 +11,7 @@ defmodule Gnat.Jetstream.API.Stream do
 
   Stream struct fields explanation:
 
+  * `:allow_direct` - Allow higher performance, direct access to get individual messages. E.g. KeyValue
   * `:allow_rollup_hdrs` - allows the use of the Nats-Rollup header to replace all contents of a stream,
     or subject in a stream, with a single new message.
   * `:deny_delete` - restricts the ability to delete messages from a stream via the API. Cannot be changed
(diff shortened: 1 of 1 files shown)


djcarpe:
Adding an issue https://github.com/nats-io/nats.ex/issues/162

djcarpe:
And finally an example using my fork that does not override fields.

# Untitled notebook

```elixir
Mix.install([
  {:gnat, git: "https://github.com/djcarpe/nats.ex", tag: "0.00"}
])
```

## Section

```elixir
{:ok, conn} = Gnat.start_link(%{host: ~c"127.0.0.1", port: 4222})

{:ok, info} = Gnat.Jetstream.API.Stream.info(conn, "test")

config = info.config |> Map.put(:max_age, 130_000_000_000)

{:ok, _} = Gnat.Jetstream.API.Stream.update(conn, config)
```

autodidaddict (approved):

mmmries (approved):
:shipit: these changes look spot on. I'm not sure why the update would have affected those fields if our code wasn't even trying to send them (maybe there's a different default for update than for create?), but either way being explicitly aware of the flags and sending the values from the user is the right thing to do. Thanks!

mmmries:
I've released this change in the 1.8.4 version https://github.com/nats-io/nats.ex/releases/tag/v1.8.4

PR #169, opened 2024-09-14
Add list_buckets function and tests to Object API

### LINKED ISSUE #168: Cannot list Jetstream object buckets
Missing functionality to list existing object buckets.

Equivalent CLI command is `nats object ls`.

Similar to https://github.com/nats-io/nats.ex/issues/152

jcreager:
This PR closes #168 and adds the ability to list Object bucket streams, replicating the nats CLI command `nats object ls`.

Credit to @lsxliron as this PR is based off of their PR to add the `list_buckets` function to the KV API.

### DIFF
--- a/lib/gnat/jetstream/api/object.ex
+++ b/lib/gnat/jetstream/api/object.ex
@@ -145,6 +145,38 @@ defmodule Gnat.Jetstream.API.Object do
     end
   end
 
+  @doc """
+  Returns true if the provided stream is an Object bucket, false otherwise
+  ## Parameters
+  * `stream_name` - the stream name to test
+  """
+  @spec is_object_bucket_stream?(stream_name :: binary()) :: boolean()
+  def is_object_bucket_stream?(stream_name) do
+    String.starts_with?(stream_name, "OBJ_")
+  end
+
+  @doc """
+  Returns a list of all Object buckets
+  """
+  @spec list_buckets(conn :: Gnat.t()) :: {:error, term()} | {:ok, list(String.t())}
+  def list_buckets(conn) do
+    with {:ok, %{streams: streams}} <- Stream.list(conn) do
+      stream_names =
+        streams
+        |> Enum.flat_map(fn bucket ->
+          if is_object_bucket_stream?(bucket) do
+             [bucket |> String.trim_leading(@stream_prefix)]
+          else
+             []
+          end
+        end)
+      {:ok, stream_names}
+    else
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
   defp stream_name(bucket_name) do
     "#{@stream_prefix}#{bucket_name}"
   end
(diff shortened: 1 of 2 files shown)


mmmries:
Hey @jcreager thank you so much for opening the issue and then putting up a PR ❤️ 💛 💙 💚 💜
On my first reading I don't see any problems here, but I'm short on time and haven't had much time to think about it. I'd like to take a second pass before making a final decision. Hopefully I can find time this week to review and leave any notes.

jcreager:
Right on, thanks @mmmries.

lib/gnat/jetstream/api/object.ex
```diff
@@ -145,6 +145,38 @@ defmodule Gnat.Jetstream.API.Object do
...
+  def is_object_bucket_stream?(stream_name) do
+    String.starts_with?(stream_name, "OBJ_")
+  end
+
+  @doc """
+  Returns a list of all Object buckets
+  """
+  @spec list_buckets(conn :: Gnat.t()) :: {:error, term()} | {:ok, list(String.t())}
+  def list_buckets(conn) do
+    with {:ok, %{streams: streams}} <- Stream.list(conn) do
```
mmmries, line 163: The `Stream.list` function accepts optional arguments to fetch additional pages of streams. It looks like both `Object.list_buckets` and `KV.list_buckets` only look for matching stream names in the first page of stream names. Maybe we should accept optional arguments to look at the next page? I'd be interested to know how this is handled by the cli.

mmmries (approved):
The only comment I left is something that would apply only for cases where Jetstream is hosting a large number of streams and we would potentially miss a bucket.

I think we can tackle that in a follow-up PR since it affects both of these functions and we'll probably want to handle them the same

lib/gnat/jetstream/api/object.ex
jcreager: @mmmries good callout.  I had not considered this.  I do agree that this can be handled in a separate PR.  I have some thoughts below on how we could accomplish this.  Let me know what you think.

CLI implements list buckets [here](https://github.com/nats-io/natscli/blob/0fc79e766be8ea67c07b13c869563f1b8e30b848/cli/object_command.go#L348).  I think that listing streams is implemented [here](https://github.com/nats-io/natscli/blob/0fc79e766be8ea67c07b13c869563f1b8e30b848/cli/stream_command.go#L748).

I can think of a couple of scenarios to consider.

* `list_buckets` in both modules adhere to the original contract of `Stream.list` (`{:ok, %{total: _, offset: 0, limit: 1024, streams: _}}`).  This would be a breaking change to KV's `list_buckets`.
* `list_buckets` for both KV and Object have their own contract that looks like `{:ok, ["TEST_BUCKET_1", "TEST_BUCKET_2"], %{total:_, offset: 0, limit: 1024}}`.  `list_buckets/2` in either module will optionally accept a parameter which specifies the offset and limit of the next page.

I think that a breaking change should be avoided.  `list_bucket/2` can expose the page API to the client and let the client handle pagination afterwards.

jcreager:
@mmmries I have created #172 to handle the pagination feature.

jcreager:
Hey @mmmries just following up to see if you feel this is ok to merge.

mmmries:
I'll get this released in a new patch version this evening. Thanks again for the contribution

jcreager:
Thanks @mmmries!  I'm happy I was able to make a contribution.

mmmries:
Released as version 1.8.5 https://hex.pm/packages/gnat/1.8.5

PR #176, opened 2024-10-16
Housecleaning

mmmries:
This PR handles some general housecleaning, making sure the project is up-to-date with the most recent versions of Elixir and Erlang and updating our testing matrix.


mmmries:
@autodidaddict it looks like the latest `nkeys` library does not build cleanly on OTP 24. I'm thinking it's time to update our compatibility and testing matrix anyway. What would you think about supporting OTP 25+ and Elixir 1.14+? I'm going to try that in this PR

Screenshot 2024-10-16 at 1 28 38 PM

autodidaddict:
I think that's a good idea. The mix file from `nkeys.ex` has a requirement for Elixir 1.17 (it comes that way with "mix new" now, apparently...), which is I think where the OTP dependency is coming from as well.

mix.lock
```diff
@@ -1,26 +1,32 @@
...
   "nimble_options": {:hex, :nimble_options, "1.0.2", "92098a74df0072ff37d0c12ace58574d26880e522c22801437151a159392270e", [:mix], [], "hexpm", "fd12a8db2021036ce12a309f26f564ec367373265b53e25403f0ee697380f1b8"},
   "nimble_parsec": {:hex, :nimble_parsec, "1.4.0", "51f9b613ea62cfa97b25ccc2c1b4216e81df970acd8e16e8d1bdc58fef21370d", [:mix], [], "hexpm", "9c565862810fb383e9838c1dd2d7d2c437b3d13b267414ba6af33e50d2d1cf28"},
-  "nkeys": {:hex, :nkeys, "0.2.1", "c183105135de9d0c07d649f6a13e39792b095772cdbeaacf37dea4e35b647100", [:mix], [{:ed25519, "~> 1.3", [hex: :ed25519, repo: "hexpm", optional: false]}], "hexpm", "946551188bd103559e5129835f782f32ea26027d0ee7063eb6910ffd707fc1aa"},
+  "nkeys": {:hex, :nkeys, "0.3.0", "837add5261a3cdd8ff75b54e0475062313093929ab5e042fa48e010f33b10d16", [:mix], [{:ed25519, "~> 1.3", [hex: :ed25519, repo: "hexpm", optional: false]}, {:kcl, "~> 1.4", [hex: :kcl, repo: "hexpm", optional: false]}], "hexpm", "b5af773a296620ee8eeb1ec6dc5b68f716386f7e53f7bda8c4ac23515823dfe4"},
```
mmmries, line 25: @autodidaddict are there new functions in the `nkeys` library that you're wanting to use in the nats client? I'm wondering if we should bump our dependency to `~> 0.3` or if we should leave it at `~> 0.2` so it's an optional upgrade for people?

mmmries:
> I think that's a good idea. The mix file from nkeys.ex has a requirement for Elixir 1.17 (it comes that way with "mix new" now, apparently...), which is I think where the OTP dependency is coming from as well.

The test matrix now uses elixir 1.14, 1.16 and 1.17 (it's defined as `main` to keep up with whatever is latest at the time of the CI job). So that definition doesn't appear to break backward compatibility for this library. I think the previous failure was because the binary pattern match didn't previously support the `-` operator. So we could bump the nkeys library to `elixir: "~> 1.14"` if we wanted to, but it it doesn't seem to be a problem for users of this library if you want to leave it as-is.

PR #179, opened 2024-12-11
Fix incorrect useage of charlist

davydog187:
Single quotes (charlists) were used instead of double quotes (binaries).

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ iex(6)> Gnat.pub(gnat, "topic", "ohai")
 ```
 
 The `pub` , `sub` , `request` , and `unsub` events all report the latency of those respective calls.
-The `message_received` event reports a number of messages like `%{count: 1}` because there isn"t a good latency metric to report. Any microservices managed by a consumer supervisor will also report `service_request` and `service_error`. In addition to the `:topic` metadata, microservices will also include `:endpoint` and `:group` (which can be `nil`) in their telemetry reports.
+The `message_received` event reports a number of messages like `%{count: 1}` because there isn't a good latency metric to report. Any microservices managed by a consumer supervisor will also report `service_request` and `service_error`. In addition to the `:topic` metadata, microservices will also include `:endpoint` and `:group` (which can be `nil`) in their telemetry reports.
 
 All of the events (except `unsub` ) include metadata with a `:topic` key so you can split your metrics by topic.
 
--- a/README.md
+++ b/README.md
@@ -15,9 +15,9 @@ The goals of the project are resiliency, performance, and ease of use.
 ## Usage
 
 ``` elixir
-{:ok, gnat} = Gnat.start_link(%{host: '127.0.0.1', port: 4222})
+{:ok, gnat} = Gnat.start_link(%{host: "127.0.0.1", port: 4222})
 # Or if the server requires TLS you can start a connection with:
-# {:ok, gnat} = Gnat.start_link(%{host: '127.0.0.1', port: 4222, tls: true})
+# {:ok, gnat} = Gnat.start_link(%{host: "127.0.0.1", port: 4222, tls: true})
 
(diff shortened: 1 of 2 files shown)


lib/gnat.ex
```diff
@@ -88,30 +88,30 @@ defmodule Gnat do
...
-    :server_id => binary(),
-    :server_name => binary(),
-    :version => binary(),
-    optional(:ldm) => boolean(),
-    optional(:tls_verify) => boolean(),
-    optional(:tls_available) => boolean(),
-    optional(:tls_required) => boolean(),
-    optional(:auth_required) => boolean(),
-  }
+          :client_id => non_neg_integer(),
```
davydog187, line 91: Looks like my formatter changed this as well. If you'd like, I can split this out into a different PR

mmmries:
Hey @davydog187 thanks for opening a PR ❤️ 💛 💙 💚 💜 

Most of these arguments do (or at least did) have a requirement to be passed a charlists because they are passthrough arguments to erlang stdlib functions (opening a TLS or TCP socket). Did you test passing in strings/binaries? Maybe the erlang functions have started to support strings?

I'm definitely open to updating the docs and improving typespecs if the underlying functions support this. It's definitely more idiomatic for Elixir libraries

README.md
```diff
@@ -49,11 +49,11 @@ end
 ## TLS Connections
 
 [NATS Server](https://github.com/nats-io/nats-server) is often configured to accept or require TLS connections.
-In order to connect to these clusters you'll want to pass some extra TLS settings to your `Gnat` connection.
+In order to connect to these clusters you"ll want to pass some extra TLS settings to your `Gnat` connection.
```
mmmries, line 52: Haha, probably didn't mean to swap out this apostraphe for a quote

lib/gnat.ex
mmmries: I thought I had a formatter check built into the CI process, but maybe I'm misremembering. If this is the standard formatter settings for current elixir versions then we should just apply it now.

README.md
```diff
@@ -134,7 +134,7 @@ Before running the tests make sure you have a locally running copy of `nats-serv
 
 We currently use version `2.6.6` in CI, but anything higher than `2.2.0` should be fine.
 Versions from `0.9.6` up to `2.2.0` should work fine for everything except header support.
-Make sure to enable jetstream with the `nats-server -js` argument and you might also want to enable debug and verbose logging if you're trying to understand the messages being sent to/from nats (ie `nats-server -js -D -V`).
+Make sure to enable jetstream with the `nats-server -js` argument and you might also want to enable debug and verbose logging if you"re trying to understand the messages being sent to/from nats (ie `nats-server -js -D -V`).
```
doomspork, line 137: Same as above:

```suggestion
Make sure to enable jetstream with the `nats-server -js` argument and you might also want to enable debug and verbose logging if you're trying to understand the messages being sent to/from nats (ie `nats-server -js -D -V`).
```

davydog187:
Cleaning up now, sorry @mmmries I was moving too fast 😅

lib/gnat.ex
davydog187: TBH i'm not sure where the difference is coming from

davydog187:
@mmmries regarding the question of binaries vs charlists, `:gen_tcp` still expects a charlist https://www.erlang.org/doc/apps/kernel/inet#t:hostname/0, but `Gnat.Handshake.connect/1` always converts it to a charlist

https://github.com/nats-io/nats.ex/blob/v1.9.0/lib/gnat/handshake.ex#L10

davydog187:
Should be good to run CI now @mmmries

mmmries:
Double thanks for taking a second pass and confirming the details on the binary vs charlists. Appreciate the contribution

PR #183, opened 2024-12-30
Soft deprecate is_kv_bucket_stream?/1 in favor of kv_bucket_stream?/1

davydog187:
the `is_` prefix is [reserved for guards](https://hexdocs.pm/elixir/1.18.1/naming-conventions.html#is_-prefix-is_foo) by convention. This PR aligns the function name with the convention for [functions that return booleans](https://hexdocs.pm/elixir/1.18.1/naming-conventions.html#trailing-question-mark-foo)


mmmries (approved):
👍 I had never seen that naming conventions doc. Thanks for linking it

PR #186, opened 2025-01-04
Remove function_exported? check for Keyword.validate!/2

davydog187:
Gnat already requires Elixir 1.14, so we're safe to remove


mmmries (approved):
:shipit:

PR #188, opened 2025-01-06
Test on Elixir 1.18 and NATS 2.10.24

davydog187:


.github/workflows/CI.yml
```diff
@@ -22,11 +22,15 @@ jobs:
...
-            nats: "2.10.14"
+            nats: "2.10.24"
 
           - otp: "27"
             elixir: "1.17"
-            nats: "2.10.14"
+            nats: "2.10.24"
+
+          - otp: "27"
+            elixir: "1.18"
```
mmmries, line 32: Won't the `main` branch block below cover 1.18.* development? I've been in the habit of adding older versions once there's a new minor version under development. Do you think it's valuable to have `1.18` called out separately from the `main`?

davydog187: Yes, it will cover development on the upcoming 1.19 release and any possible changes.

With that being said, I wouldn't personally test against main and would drop it in favor of the explicit 1.18

davydog187: To clarify my position here, I would advise testing against "nightly" builds such as main here, since they tend to have bugs and may introduce intermediate failures to CI.

For a library, it's better to keep the main CI against the latest stable releases.

If you want to test against main, I would do that outside of the PR cycle, possibly as a nightly test.

mmmries: I think you're right. The `main` branch is probably too unstable to be part of our CI. I'll just remove it for now as we have a schedule monthly build which would check for all of the supported minor versions

Issue #172, opened 2024-09-18
Support pagination for KV and Object `list_bucket`


jcreager:
Proposal:

`list_buckets` for both KV and Object have their own contract that looks like {:ok, ["TEST_BUCKET_1", "TEST_BUCKET_2"], %{total:_, offset: 0, limit: 1024}}. 

`list_buckets/2` in either module will optionally accept a parameter which specifies the offset and limit of the next page.

This is to avoid introducing breaking changes to the existing implementation of KV's `list_object` function, as well as the proposed Object `list_buckets` function.

mmmries:
I was thinking about this a bit more and one way we could keep the API the same is to handle pagination inside the function itself. That seems to be what this function in go does: https://github.com/nats-io/nats.go/blob/f0b1ea421c52eee17f89d528543b7affba18910b/kv.go#L1150-L1167

It paginates through all the streams and for each of them that match the right prefix, it generates a KeyValueStoreStatus. We're just returning the names, but I think we could handle the pagination and then just return the full list of KV bucket names?

I also couldn't find any pagination options that are available on the CLI, so I'm guessing it just paginates through all of them as well?

PR #189, opened 2025-01-06
Gnat.Jetstream.API.KV.info/3

davydog187:
Adds a convenience function for getting info of buckets.

As a note, I think that `Stream.info/3` should be changed to take a keyword list. We can support the case where its nil/binary still

### DIFF
--- a/test/jetstream/api/kv_test.exs
+++ b/test/jetstream/api/kv_test.exs
@@ -155,6 +155,7 @@ defmodule Gnat.Jetstream.API.KVTest do
       :ok = KV.delete_bucket(:gnat, bucket)
     end
   end
+
   describe "list_buckets/2" do
     test "list buckets when none exists" do
       assert {:ok, []} = KV.list_buckets(:gnat)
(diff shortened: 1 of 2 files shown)


test/jetstream/api/kv_test.exs
```diff
@@ -170,10 +171,23 @@ defmodule Gnat.Jetstream.API.KVTest do
...
       assert {:ok, _response} = Stream.create(:gnat, stream)
       assert {:ok, ["TEST_BUCKET_1"]} = KV.list_buckets(:gnat)
       :ok = KV.delete_bucket(:gnat, "TEST_BUCKET_1")
     end
   end
+
+  describe "info/3" do
+    test "returns bucket info" do
+      assert {:ok, _} = KV.create_bucket(:gnat, "TEST_BUCKET_1")
+      assert {:ok, %{config: %{name: "KV_TEST_BUCKET_1"}}} = KV.info(:gnat, "TEST_BUCKET_1")
```
davydog187, line 189: delete this bucket!

mmmries (approved):
For the keyword list, do you mean that rather than passing just an optional `domain` as the third argument, we would accept a keyword list like `Stream.info(conn, name, domain: "foo")`?

I like that idea just to keep things as consistent as possible across the library

mmmries:
We also still need to deal with #172 on these functions. But I don't think it's super common for people to have more than a page worth of streams in their deployment, or at least not in a place where they are using this library to list all the KV buckets

PR #191, opened 2025-03-06
make KV-watcher emit `:key_added` events when the message has a header

rixmann:
When creating keys (or updating) with the `{"nats-expected-last-subject-sequence", "<seq>"}` header, KV watcher did not trigger the handler.

### DIFF
--- a/lib/gnat/jetstream/api/kv/watcher.ex
+++ b/lib/gnat/jetstream/api/kv/watcher.ex
@@ -12,6 +12,7 @@ defmodule Gnat.Jetstream.API.KV.Watcher do
 
   @operation_header "kv-operation"
   @operation_del "DEL"
+  @operation_purge "PURGE"
 
   @type keywatch_handler ::
           (action :: :key_deleted | :key_added, key :: String.t(), value :: any() -> nil)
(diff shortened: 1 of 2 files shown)


lib/gnat/jetstream/api/kv/watcher.ex
```diff
@@ -56,6 +56,8 @@ defmodule Gnat.Jetstream.API.KV.Watcher do
 
     if {@operation_header, @operation_del} in headers do
       state.handler.(:key_deleted, key, body)
+    else
```
mtmk, line 59: not sure if it's applicable here but there is also a `purge` operation. assuming it's an add if not deleted might not be right.

rixmann: tested with purge and adapted, purge also triggers `:key_deleted` now.
updates will still show up as `:key_added` and i don't see a way to differentiate between added and updated without
inspecting the history.

rixmann: ok, done

mtmk (approved):
LGTM thanks @rixmann sorry for the delay

tested locally, works as expected.

mmmries (approved):
Code looks reasonable to me. Thanks @mtmk for reviewing with some Jetstream expertise. Glad this matches the community expectations.

PR #192, opened 2025-03-19
add `:compression` to stream attributes

rixmann:


test/jetstream/api/stream_test.exs
```diff
@@ -144,7 +144,7 @@ defmodule Gnat.Jetstream.API.StreamTest do
     assert {:ok, _response} = Stream.create(:gnat, stream)
 
     assert {:ok, response} = Stream.info(:gnat, "INFO_TEST")
-    assert response.config == stream
+    assert response.config == stream |> Map.put(:compression, "none")
```
mtmk, line 147: why do we need this additional `Map.put`?

mtmk (commented):
Looks good in general, just a question about test

test/jetstream/api/stream_test.exs
rixmann: you are right, i overlooked this.
initially i thought in order to get no compression i need to send the empty string - which didn't work - so i had no default for "compression" in the struct.
But then i saw that "none" worked and is even part of the stream info - so i changed compression to use "none" as default (which is much simpler for the struct constrains).
=> i will remove the Map.put

mtmk (approved):
LGTM thanks @rixmann

mtmk:
tested locally, works as expected.

_(Not specifically related to this change)_ note: there is a general question about not sending the defaults in JSON requests e.g. when a stream create request is made from the natscli the `compression` key is not sent at all.

mmmries:
@mtmk I would love to touch base and discuss a bit more about whether to send keys with default values. Since we use structs within the library, we lose details on whether the user intended to set a key to a specific value or just left it at the default. I wonder if we need to handle cases where the default value changes in the server? If so we would need to remove the usage of structs and let users pass in something like a keyword list or a map.

I think there are better solutions than what we have, but we would need to think carefully about backward compatibility for users.

mmmries (approved):
Thanks for getting this added. Looks like a nice feature in Jetstream

PR #193, opened 2025-04-11
Hint `:timeout` option in KV's typespec

Jurshsmith:
This PR does the following:

- Includes Elixir modules and scripts in formatter configuration, allowing extensions using Elixir Dialyzer/LSP for autoformatting to find Elixir scripts and modules in a given context/directory

- Updates the type spec for KV's opts to implicit `keyword` while hinting at the potential option for configuring 
the request's `timeout`


mmmries (approved):
:shipit: Sorry for the slow response on this @Jurshsmith 

I haven't been watching this repo very closely, but this is a solid addition. Appreciate you taking the time to submit it

PR #197, opened 2025-06-27
feat: add jetstream message metadata helper

c0deaddict:
The nats python library has a useful helper function/property `message.metadata` that decodes information in the `reply_to` address of a Jetstream message: https://github.com/nats-io/nats.py/blob/d9f24b4beae541b7723873ba0a786ea7c0ecb3d5/nats/aio/msg.py#L160

This PR adds a helper function to extract the metadata information.

API doc: https://docs.nats.io/reference/reference-protocols/nats_api_reference#acknowledging-messages

### DIFF
--- a/lib/gnat/jetstream/api/message.ex
+++ b/lib/gnat/jetstream/api/message.ex
@@ -1,0 +1,104 @@
+defmodule Gnat.Jetstream.API.Message do
+  @moduledoc false
+
+  # Based on:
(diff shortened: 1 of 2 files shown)


mmmries:
❤️ 💛 💙 💚 💜 this is super awesome! I was just wondering the other day how I could keep track of consumer sequence vs stream sequence for some of my consumers and this will work super well.

I'll take a look at the code details to see if we can get this merged quickly. I'm on vacation now, but hopefully within a week or so I can find time to spend some time on it

mmmries (approved):
Looks good. Glad to see that it handles both the 12 part and 9 part reply-to values

PR #198, opened 2025-06-29
use Pager module for KV.contents

### LINKED ISSUE #182: Speed up the implementation of NATS.Jetstream.API.KV.contents/2
## Problem
While testing the NATS KV store, I was surprised to find out that reading the contents of a bucket was an order of magnitude slower than reading an individual key.

Since its an aggregate operation, I did expect it to be slower, but not by this significant of a margin.

```elixir
iex(1)> Gnat.Jetstream.API.KV.put_value(:cluster_nats, "example", "foo", "bar")
:ok

# Reading a value single digit milliseconds
iex(2)> :timer.tc(fn -> Gnat.Jetstream.API.KV.get_value(:cluster_nats, "example", "foo") end, :millisecond)
{4, "bar"}

# Reading the contents of a bucket is always at least 100ms
iex(3)> :timer.tc(fn -> Gnat.Jetstream.API.KV.contents(:cluster_nats, "example") end, :millisecond)
{134, {:ok, %{"foo" => "bar"}}}
```

Upon further inspection, it appears that this is a limitation of the client implementation, where it will always wait a minimum of 100ms from the consumer before returning https://github.com/nats-io/nats.ex/blob/v1.9.0/lib/gnat/jetstream/api/kv.ex#L263

## Possible Solutions
1. Document this behavior and make the timeout configurable, noting the risk of missing keys if your timeout is too short
2. Explore other protocol-level features for an alternative implementation (I'm not familiar enough yet to make a determination)
3. Investigate implementations in other languages and copy their strategy
4. ???

I am happy to submit a PR but I wanted to start a conversation first

mmmries:
Fixes #182 

/cc @davydog187 @brandynbennett 

I had a use-case come up at work where I wanted to create a small stream to track which "migrations" had already been applied to jetstream. This way application engineers could define their own migrations and we could apply any pending migrations as part of a code deployment automatically.

While working on that I remembered about the `Pager` module which I'm still leaving undocumented for now in this PR, but it seems to be pretty stable and now accepts `first_seq` as an optional argument. I re-used that for the implementation of `KV.contents` which essentially just reduces over the KV stream and builds up a map of the contents.

This removes the need for a push-based consumer with a 100ms timeout. We can request batches of messages up until the server tells us that there are no more. In my local testing, this reduced the time for the `KV.contents` function from ~108ms to ~6-7ms


lib/gnat/jetstream/pager.ex
```diff
@@ -8,52 +8,49 @@ defmodule Gnat.Jetstream.Pager do
...
       receive_messages(state, [])
     end
   end
 
   def cleanup(%{conn: conn} = state) do
-    with :ok <- Gnat.unsub(conn, state.sub),
-         :ok <- Consumer.delete(conn, state.stream_name, state.consumer_name, state.domain) do
-      :ok
+    with :ok <- Gnat.unsub(conn, state.sub) do
+      :ok = Consumer.delete(conn, state.stream_name, state.consumer_name, state.domain)
```
davydog187, line 53: This will crash now with bad match, is that intended? Or does this always return ok?

mmmries: It's possible to get an `{:error, "some message"}` response from this API, but I don't think I would really add any error handling for that case. Since we're deleting the consumer that we created at the beginning of this same process, there isn't much that we could try to do to recover from this error. So I left it as something that would raise an exception if it failed.

PR #200, opened 2025-06-30
implement regularly scheduled PING/PONG health check

### LINKED ISSUE #51: Detect network failure
I think it would be nice to have a quick way to detect when the network connection disappears unexpectedly. I'm thinking some kind of inactivity timeout that is renewed every time we receive data from the server. When the inactivity timeout is exceeded, a ping should be sent. If the ping times out the GenServer should be stopped.

Not sure how best to implement. Could it be another process that has something like:

```elixir
# some other module ?
def activity_monitor
  receive do
    :refresh -> activity_monitor
    after 5_000 -> Gnat.ping_or_stop_async(state.gnat)
  end
end

# Gnat module
def handle_info({:tcp, socket, data}, %{socket: socket, parser: parser}=state) do
  ...
  ActivityMonitor.call(state.activity_monitor, :refresh)
  ...
end
```

I know that is not a fully fleshed out idea, but I think it's enough to understand how I'm thinking about implementing it.

mmmries:
Fixes #51 

This PR implements a regularly scheduled PING/PONG health check. By default we will run this every 10sec (configurable by the user via the `ping_interval` option). After a successful handhake, we will send a `PING` at each interval and we set a state of `waiting_on_pong: true`. This state is cleared whenever we receive a `PONG` from the server (in my testing with a local NATS server, this always came back in less than 1ms).

If we get to the next ping interval and we are still waiting on a PONG, we crash the process with an error message about not having received a PONG.

### DIFF
--- a/lib/gnat.ex
+++ b/lib/gnat.ex
@@ -43,6 +43,7 @@ defmodule Gnat do
   @typedoc """
   * `connection_timeout` - limits how long it can take to establish a connection to a server
   * `host` - The location of the NATS server
+  * `ping_interval` - The number of milliseconds between sending PING messages to the server to check the health of our connection
   * `port` - The port the NATS server is listening on
   * `ssl_opts` - Options for connecting over SSL
   * `tcp_opts` - Options for connecting over TCP
@@ -54,6 +55,7 @@ defmodule Gnat do
           optional(:connection_timeout) => non_neg_integer(),
           optional(:host) => binary(),
           optional(:inbox_prefix) => binary(),
+          optional(:ping_interval) => non_neg_integer(),
           optional(:port) => non_neg_integer(),
           optional(:ssl_opts) => list(),
           optional(:tcp_opts) => list(),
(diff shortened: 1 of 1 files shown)


mmmries:
I pushed a hex release called `1.10.2-rc0` from this branch, so I can test running this on my raspberry pi project that has been sporadically running into these dead connection issues.

mmmries:
I tested our the `rc0` on a nerves device and found that I was getting server `INFO` messages after each `PING/PONG` interaction and `gnat.ex` was not handling those (I assumed we only got INFO messages on handshake, but it looks like these are regular updates that give the client fresh information about the `connect_urls` for handling a reconnect)

I added a commit to update the `server_info` when we get additional INFO messages, but I'll have to think about how we can make use of this additional information for things like the `ConnectionSupervisor` so we can keep an updated list of the potential reconnect addresses

mmmries:
After letting this bake for a full day (connected to the NGS public NATS cloud) and passing all of our automated checks, I feel good about this PR.

PR #217, opened 2026-03-19
Use flow_control for Object.get

ppkn:
We were experiencing issues with "Slow Consumer" messages when using Object.get with a combination of large files (~100MB) and high-latency chunk functions like file and network IO. The nats server was cutting off the consumer before it could process the whole file.

Digging into nats.go I discovered that they use the flow_control setting for their Object.get consumer. This requires sending ack replies to FlowControl messages from the server. The server uses this to control how much data it sends to not overwhelm the consumer.

This PR solves our problem internally and will bring the library more in line with the nats.go implementation.

I looked into using the Pager module but it doesn't fit well because we need to pass the connection to each call in case we need to reply to a FlowControl message. That didn't fit well with the current implementation of Pager.reduce/5

### DIFF
--- a/lib/gnat/jetstream/api/object.ex
+++ b/lib/gnat/jetstream/api/object.ex
@@ -284,24 +284,37 @@ defmodule Gnat.Jetstream.API.Object do
         ack_policy: :none,
         max_ack_pending: nil,
         replay_policy: :instant,
-        max_deliver: 1
+        max_deliver: 1,
+        flow_control: true,
+        idle_heartbeat: 5_000_000_000
       })
 
-    :ok = receive_chunks(sub, meta.chunks, chunk_fun)
+    :ok = receive_chunks(conn, sub, meta.chunks, chunk_fun)
 
     :ok = Gnat.unsub(conn, sub)
     :ok = Consumer.delete(conn, stream, consumer.name)
   end
 
-  defp receive_chunks(_sub, 0, _chunk_fun) do
+  defp receive_chunks(_conn, _sub, 0, _chunk_fun) do
     :ok
   end
 
-  defp receive_chunks(sub, remaining, chunk_fun) do
+  defp receive_chunks(conn, sub, remaining, chunk_fun) do
     receive do
+      # Flow control message with reply - respond to it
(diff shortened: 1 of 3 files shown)


lib/gnat/jetstream/api/object.ex
```diff
@@ -284,24 +284,37 @@ defmodule Gnat.Jetstream.API.Object do
...
-  defp receive_chunks(_sub, 0, _chunk_fun) do
+  defp receive_chunks(_conn, _sub, 0, _chunk_fun) do
     :ok
   end
 
-  defp receive_chunks(sub, remaining, chunk_fun) do
+  defp receive_chunks(conn, sub, remaining, chunk_fun) do
     receive do
+      # Flow control message with reply - respond to it
+      {:msg, %{sid: ^sub, status: "100", description: "FlowControl Request", reply_to: reply}}
```
ppkn, line 305: This pattern match and the `status: "100"` one below are based on information found [in the docs](https://docs.nats.io/reference/reference-protocols/nats_api_reference#acknowledging-messages). I also checked nats.go to see if there were other FlowControl messages but this seems to be [the only one](https://github.com/nats-io/nats.go/blob/9eec10ce34808e1a0949f9b8f1e968c1e6484b8d/jetstream/message.go#L152).

test/jetstream/api/object_test.exs
```diff
@@ -214,15 +235,15 @@ defmodule Gnat.Jetstream.API.ObjectTest do
     end
   end
 
-  # create a random 1MB binary file
+  # create a random 100MB binary file
```
ppkn, line 238: This change didn't noticeably increase the spec run time locally.

ppkn (commented):
Hopefully I'm not over-commenting the cases in the receive block. Let me know if there is a more idiomatic way of doing this.

mmmries (approved):
:shipit: reviewed this and I was wondering if we should expose more of the `Pager` module so we could implement this function using a pull consumer (with natural backpressure) rather than using a push consumer + flow_control, but this matches what the go client does and it's a valid approach with strong support from the community and the server.

Refactoring the `Pager` would be a bigger change and require a more thorough review, so I think this solution is good.

mmmries:
this is included in the 1.13.1 release

PR #218, opened 2026-04-01
introduce on_connected callback

mmmries:
I'm working on a use-case at work where I want to subscribe to a KV bucket and I wanted to have the ability to detect when the data has been initially loaded and we're now just waiting on future updates. In this case each running instance has it's own ephemeral consumer for a relatively small stream and it's really useful to know how many pending messages are outstanding at the time that my PullConsumer starts up.

I can also check on progress by checking the metadata on the subsequent `handle_message` function calls, but if the KV bucket is empty at startup, I need to know that there are 0 pending messages in order to mark that my consumer has "caught up" and is ready to be used.

### DIFF
--- a/lib/gnat/jetstream/pull_consumer.ex
+++ b/lib/gnat/jetstream/pull_consumer.ex
@@ -261,6 +261,42 @@ defmodule Gnat.Jetstream.PullConsumer do
               {ack_action, new_state}
             when ack_action: :ack | :nack | :term | :noreply, new_state: term()
 
+  @doc """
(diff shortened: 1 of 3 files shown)


mmmries:
Going to merge this into the branch where I'm preparing a 1.14.0.rc release

0xAX:
@mmmries looks good to me although I am late a bit :) In general thanks for the change, I have something similar implemented in app level, nice to have it in upstream

djcarpe:
Sorry! I’m in the middle of moving, but my cursory look it seems like a sensible change. I’ll give it a better look when I’m settled.

PR #225, opened 2026-05-27
Fix large payload parsing performance

cbecker:
First, thanks for creating this great library for Elixir!

### Why this PR

We need to handle large payloads (> 8MB), and we noticed a significant lag as the payload increased.

*Before this PR*
```
1 MB  payload:  21.86 ms latency
16 MB payload: 2.6 seconds latency
32 MB payload: 9.9 seconds latency
```

*After this PR*
```
1 MB  payload:  2.54 ms latency
16 MB payload: 15.05 ms latency
32 MB payload: 29 ms latency
```

'latency': delta t between pushing to NATS server and receiving the message back in the subscribed process' inbox.

### Explanation

When a NATS message body spans multiple TCP segments, the parser was
accumulating data with `partial <> string` on every incoming chunk.
This means a lot of copies, and the more chunks, the more copies needed.

### Fix
This PR introduces a `pending` field on `Parsec` that tracks the
already-decoded MSG/HMSG header metadata and body bytes received so
far as an iolist. Subsequent chunks are appended to the iolist in O(1)
with no copying. `iolist_to_binary/1` is called exactly once, only
when enough bytes have arrived to complete the message body.

This reduces large-payload receive latency from O(n²) to O(n) and
eliminates the main source of multi-second delays observed with
payloads above ~8 MB.

### Benchmarks
I ran all benchmarks but the client/server ones, and saw improvements overall.
I am happy to provide more details, but I am not sure what's the best way to compare or provide before-after.
I'm glad to receive feedback and provide more details.

Thanks!

(Co-authored with Cursor)

### DIFF
--- a/lib/gnat/parsec.ex
+++ b/lib/gnat/parsec.ex
@@ -174,6 +174,8 @@ defmodule Gnat.Parsec do
   # Accumulate incoming TCP chunks into an iolist until enough bytes have arrived for the
   # full message body (plus the trailing \r\n). Returns {:accumulating, chunks, have} when
   # more data is still needed, or {:complete, binary} when ready to extract the body.
+  @spec accumulate_body(iolist(), non_neg_integer(), binary(), non_neg_integer()) ::
+          {:accumulating, iolist(), non_neg_integer()} | {:complete, binary()}
   defp accumulate_body(chunks, have, string, needed_length) do
     new_have = have + byte_size(string)
     new_chunks = [chunks, string]
--- a/lib/gnat/parsec.ex
+++ b/lib/gnat/parsec.ex
@@ -99,7 +99,21 @@ defmodule Gnat.Parsec do
 
   def new, do: %__MODULE__{}
 
-  # Fast path: we already parsed the MSG header and are just waiting for the body bytes.
+  # Hot path: no partial header fragment, no pending body. This is the common case for
+  # small messages that arrive complete in a single TCP chunk.
+  def parse(%__MODULE__{partial: nil, pending: nil} = state, string) do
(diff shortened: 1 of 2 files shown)


mmmries:
@cbecker thank you for raising this and for the excellent PR description. It makes the review process really nice. I have a few thoughts after an initial review:

__Avoid the hard match on `"\r\n"`__

In `parsec:129` and `parsec:152` there's a straight pattern match on the final couple of bytes which is pretty safe, but the existing parsec code has `case` clauses for these kinds of matches and returns and `:error` tuple. I think it would be nice to do something similar here and return an error tuple calling out that we got a malformed message termination. This would make troubleshooting/debugging a lot easier if it comes up.

__Test the HMSG Path__

Your test changes are clean, but I think we should have a test case that covers both the MSG and HMSG path that mimics a multi-packet parsing process. This helps to document the intended behavior and gives us a nice unit-testable coverage on your change.

__Extract a shared helper for MSG and HMSG__

The new function matches you added for these two cases look very similar. Do you think it would be worth splitting out a shared helper for them? I haven't thought enough about this to be convinced one way or the other, just something I noticed.

Are those changes something you're open to making? If so I think we can get this merged and released pretty quickly.

cbecker:
Thanks @mmmries for the quick feedback. I just pushed a few commits addressing your comments, including tests for the error handling when "\r\n" are not matched.

Let me know if this is what you had in mind.

EDIT: I now also removed the last argument of `finish_msg()` that wasn't needed anymore

mmmries:
This change has been released as version 1.15.1

Thanks again for the thorough investigation and focused fix

cbecker:
Thanks for the quick review and release!

PR #226, opened 2026-07-01
refacotr(headers): gets rid of cowlib dep

tank-bohr:
cowlib dependency was introduced in https://github.com/nats-io/nats.ex/pull/116 to support headers and has pretty limited usage in the library (just for encoding/decoding headers)
At the same time the cowlib brings some frictions. It has the reported CVE which is flagged on https://hex.pm/packages/cowlib/advisories and is rejected to be fixed by the author (see [discussion](https://github.com/ninenines/cowlib/issues/152#issuecomment-4485676579))

> This CVE will likely remain as "won't fix". Expect a few more like this in the future.

And this could be a problem. For example, since https://github.com/hexpm/hex/pull/1150 `mix hex.audit` fails the task on known CVE's and there is no way to configure it to suppress it.


mmmries:
Thank you very much for this contribution. I would definitely like to remove the dependency. I'll review in depth and compare the dependency solution vs your proposed solution. In the meantime, can you give me any more context on how your parsing solution compares to cowlib? Is this the equivalent of what cowlib does? We have a method of confirming what kinds of header encodings are possible from all known versions of the gnat server so we can test exhaustively?

One of my big concerns here is making sure that we don't introduce a breaking change, or a major performance degradation for existing users of the library.

tank-bohr:
Sure. Cowboy uses a pure erlang with pattern matching for parsing http protocol. It's a bit unusual. The vast majority of the libraries both in erlang and elixir use builtin function from the OTP - [`:erlang.decode_packet/3.`](https://www.erlang.org/doc/apps/erts/erlang.html#decode_packet/3) The most popular ones such as mint and bandit are using exactly `:erlang.decode_packet/3.` It potentially could be incompatible with the cowlib. But since both approaches implement the same protocol they must be compatible.
However there are couple of cowlib-specific tricks. Cowlib (1) trims the header value and (2) lowercases the header name. Both are taken into account. The unit-tests already contain some tricky headers in test cases and they pass. But I can provide a bigger corpus and test two approaches to be identical

tank-bohr:
@mmmries Please take a look https://gist.github.com/tank-bohr/5471ffa55eab2709e737dde7457f643e

There are some discrepancies on malformed and invalid data. The valid data looks correct

mmmries:
Spent some more time spelunking on this today and I think this is a solid change. Thanks for providing the extra fuzz testing and validation. I agree that we wouldn't expect the NATS server to ever send invalid non-conformant headers, which are the only area where behavior would diverge.

❤️ 💛 💙 💚 💜 Thanks for noticing this and contributing a fig

PR #227, opened 2026-07-10
Add client name to NATS CONNECT settings

ahamez:
Hello, 

This PR adds the support of the [name parameter](https://docs.nats.io/reference/reference-protocols/nats-protocol#connect) when sending CONNECT.

I took the opportunity to silence a few warnings about variables pinning in pattern matching.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -19,10 +19,6 @@ jobs:
     strategy:
       matrix:
         pair:
-          - otp: "25"
-            elixir: "1.14"
-            nats: "2.10.24"
-
           - otp: "26"
             elixir: "1.16"
             nats: "2.10.24"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 1.16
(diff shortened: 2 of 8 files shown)


mmmries:
Hey @ahamez thanks for the contribution. It looks like pinning the sizes is actually breaking compilation on older Elixir versions. I think the name change looks solid, but can you remove the pinning from this PR? Or was there a strong reason you wanted to include it?

mmmries:
I dug a bit deeper and realized that pinning these sizes has been supported since Elixir 1.15, so I checked on the level of support for Elixir 1.14 and 1.15 and it turns out that the Elixir project itself is no longer making security patches for these versions.

So I am bumping the minimum supported version to 1.16. I then also found a tricky bug where I had an old unit test that made the assumption that the `CHANGELOG.md` file was smaller than the `README.md` which is no longer true with this small version bump 😱 

So I had to update this test and now we have a clean test suite and your change was good all along. Thanks again for the contribution

mmmries:
Released as version 1.16.0 https://hex.pm/packages/gnat/1.16.0

PR #230, opened 2026-09-21
Fix per-message acknowledgements for batch pull consumers

ericmj:
Batch pull consumers require `AckAll` and acknowledge only the last message in each batch. When two application processes share a JetStream consumer, one process can acknowledge messages the other hasn't finished processing. For example, acknowledging message 4 also acknowledges messages 1 and 2 while another process is still handling message 1. Batch mode also treats `:nack`, `:term`, and `:noreply` as successful acknowledgements.

Require `ack_policy: :explicit` for `batch_size > 1` and apply each handler result to its own message. Each successful callback is acknowledged before processing the next message. Batch consumers configured with `:all` or `:none` are rejected.

During reconnect, unsubscribe the old inbox before processing buffered messages so negative acknowledgements can't send redeliveries to an inbox being abandoned. Process the entire buffer with best-effort acknowledgements on the original connection, preserving callback state when acknowledgement sends fail. Handler failures still propagate to supervision.

Regression tests cover shared consumers, all four handler outcomes in full and partial batches, handler failures, and buffered messages across reconnects, including ephemeral consumers and consumers with `max_deliver: 1`. The default suite passed 225 tests and 19 doctests on Elixir 1.20.3/OTP 29 with NATS 2.15.0, with 16 cases excluded by repository configuration. Formatting passed.

Separate follow-up work is needed for ephemeral consumer recreation, watchdog timing during slow callbacks, and received messages still in the process mailbox when the subscription changes.


mmmries:
Thanks for the PR. I'll take a look tonight and probably again tommorrow. This would be a breaking change so I want to think carefully about different use-cases.

mmmries:
Reviewing this, I remembered that when I introduced `batch_size`, I was addressing the case of having a `PullConsumer` that subscribes to a KV stream. This is usually a case where you define an ephemeral consumer per-process (operating-system process). This gives a low-latency way to do in-memory checks for data sets that fit in memory. In this case, there will only over be one connection that consumes for the pull consumer, so you don't have to worry about consumer A acknowledeing consumer B. I think this type of pattern is the reason the server supports `ack_policy: all`.

So I'm thinking it would be great for us to support both `ack_policy: explicit` as well, but I don't want to remove the support for `ack_policy: all`.

Does that make sense to you? I feel like this was documented clearly to users wouldn't be surprised by the behavior, but if it surprised you, maybe we should call it out more clearly? And we could branch off the `ack_policy` to either support the multiple clients or the case of ephemeral consumers.

ericmj:
I've restored `:all` support in [506086e](https://github.com/ericmj/nats.ex/commit/506086e526b08abe950d755e828c1123da5cb810). Your KV cache use case can keep acknowledging the last message after every callback in the batch returns `:ack`. Batch mode also supports `:explicit`, which acknowledges each message independently and allows workers to share a consumer.

There's still a behavior change for `:all`: returning `:nack`, `:term`, or `:noreply` now stops the worker with an error and leaves the batch unacknowledged. A later cumulative acknowledgement would otherwise acknowledge the message that requested a different outcome. I've documented that restriction and the risk of sharing an `:all` consumer across workers, with regression tests for both policies and partial batches during reconnect.

mmmries:
Thanks for this great contribution. The `explicit` mode is definitely a nice improvement for durable consumers. Appreciate the quick follow-up.

PR #231, opened 2026-09-21
Fix Object Store upload validation and read integrity

ericmj:
Object Store uploads can report success when JetStream rejects chunks or metadata, and they always use 128 KiB chunks even when the bucket or connection advertises a lower limit. The first commit caps chunk size at those limits and validates each publish acknowledgement, including its stream and sequence. Server errors and malformed acknowledgements now stop the upload.

The second commit verifies streamed reads against the metadata's byte count and SHA-256 digest, adds a configurable receive deadline that heartbeats and flow control can't extend, and cleans up subscriptions and consumers on success, errors, and callback failures. Metadata listing uses the same deadline and cleanup handling. Malformed metadata returns an error. Callbacks receive data before the final integrity check, so callers must wait for `:ok` before treating the object as verified; the API documentation states this.

Added 16 Object Store regressions covering chunk limits, rejected and malformed acknowledgements, digest and size mismatches, empty objects, missing chunks, heartbeats, failed consumer creation, malformed metadata, and callback failures. On macOS with Elixir 1.20.3, OTP 29, and NATS 2.15.0, the default suite passed 227 tests and 19 doctests, with 16 cases excluded by the repository configuration. Formatting and Dialyzer passed. The full CI version matrix hasn't been run locally.


lib/gnat/jetstream/api/object.ex
```diff
@@ -252,85 +285,162 @@ defmodule Gnat.Jetstream.API.Object do
...
     {:ok, messages}
   end
 
-  defp receive_all_metas(sid, remaining, messages) do
-    receive do
-      {:msg, %{sid: ^sid, body: body}} ->
-        meta = json_to_meta(body)
-        receive_all_metas(sid, remaining - 1, [meta | messages])
-    after
-      10_000 ->
```
mmmries: Previously this 10sec timeout was a per-chunk timeout. So it would reset itself along the way. This is useful for large files that need to store many individual chunks. The new `timeout` option is a total limit. From what I know of other language clients for NATS, many of them use per-message timeouts (or per activity more generally).

I can see the total timeout being useful for clients, so maybe we should still have both? Would it make sense to derive an activity timeout based on the total? it would be weird to enforce a 10sec activity timeout, if the overall timeout is just 5sec?

Also, I think we should treat the overall timeout as optional since it's a new-new change in the library. I don't want current users to break their current use-cases if they have large objects that take more than 10sec to complete.

mmmries:
This looks like a big win for the project. Thank you for taking the time to write it and submit it 💚 💙 💛 💜 ❤️ 

It looks like the older nats server versions are running into errors on the test suite with these changes. I haven't updated the minimum server version that we support in a while, so I would be open to raising that if we think that the `2.10.X` family is just too old at this point? Or maybe we want to gate this new behavior on whether the server supports it?

I also had one concern about introducing the `timeout` in a way that is backward incompatible for current users. Totally open to hearing your feedback about that if you have a different point of view.

Thanks again for opening this, definitely something I want to get merged

ericmj:
I've updated this in [09aa875](https://github.com/ericmj/nats.ex/commit/09aa875b1dead3e5ac068acf7111b623a67205ff). `:timeout` now defaults to a 10-second wait between data chunks, and the optional `:total_timeout` defaults to `:infinity`. Each wait uses whichever limit expires first, and heartbeats and flow-control messages don't reset it.

The older-server failures came from my tests trying to change `max_consumers` on an existing stream. I've changed them to set the limit when creating the stream and added cleanup that runs even when an assertion fails. We don't need to raise the minimum server version for this.

mmmries:
Thanks for the refinement on `timeout` vs `total_timeout`. The stronger validations on both reading and writing objects will definitely help the community. Thanks @ericmj !

